PR #3, opened 2017-12-26
BFG-style Vertexcache

fholger:


renderer/BufferObject.cpp
```diff
@@ -0,0 +1,221 @@
...
+	qglBindBufferARB( GL_ARRAY_BUFFER_ARB, 0 );
+	qglBindBufferARB( GL_ELEMENT_ARRAY_BUFFER_ARB, 0 );
+}
+
+/*
+========================
+CopyBuffer
+========================
+*/
+#ifdef ID_WIN_X86_SSE2_INTRIN
```
stgatilov: We use `__SSE2__` macro for checking for SSE2 compile-time support.
It is builtin on GCC, and I have added support for it on MSVC.
Could you please replace the macro?

fholger: Absolutely. Or maybe I'll even just remove that part of code. Haven't been able to see any advantages of it in profiling...

renderer/BufferObject.cpp
```diff
@@ -0,0 +1,221 @@
...
+========================
+*/
+void * BufferObject::MapBuffer( int mapOffset ) {
+	assert( bufferObject != 0 );
+	assert( IsMapped() == false );
+
+	void *buffer = NULL;
+
+	qglBindBufferARB( bufferType, bufferObject );
+	// orphan old contents
```
stgatilov: We need to draw some VBO contents, and in the process fill another VBO, which would be drawn later.
As far as I understand, orphaning VBO is one way to achieve it (orphaned data is being read from, while you can fill the current VBO data). Another approach is to have several (e.g. two) VBOs and and switch between them: one is filled, another is being drawn from.
It seems from the code that both ways are used. Not that I understand well what's going on, but isn't it a sign that something is wrong?...

EDIT: another approach is to have one buffer, but map a portion of it while drawing from another part of it. Like a circular buffer. I guess it is rarely used because it is usually not supported by API =)

fholger: Your last point would require persistent mapping, which is an OpenGL 4.2 feature :)

But the problem here is that frontend and backend work in parallel. So if backend is rendering from buffer A and frontend is writing to buffer B, that's fine. But if you now switch the buffers and have the frontend write to buffer A without orphaning, there is a good chance that graphics card is still rendering from it. OpenGL is asynchronous, after all. So if you don't want to orphan buffer A, you'll have to wait until all relevant draw calls are truly finished, which causes stalls. Alternatively, we could use triple buffers with syncing, where stalls should virtually never happen. But they do require an additional buffer.

stgatilov: So you say that three buffers are needed to avoid stalls. That's because frontend is filling buffer for the next frame, backend is sending draw commands for the current frame, and GPU is rendering data for the previous frame, right?
I don't see orphaning in BFG code. I think explicit double/triple buffering gives more control over how much RAM is used (orphaning makes it unlimited/driver-dependent). Also, is it much less dependent on driver implementation (with orphaning you hope that driver supports the concept well).
For example, isn't it true that the current way with orphaning uses at least 4 buffers? There are two VBOs, which are often swapped, and each of them regularly gets an orphaned version also. Does GPU allocate/deallocate those orphaned arrays immediately, or it simply increases number of orphaned versions until no one wants more?

fholger: Well, BFG says GL_MAP_INVALIDATE_RANGE_BIT when mapping which supposedly should also orphan the range when needed, but it seems to not work correctly with some drivers. The reason why BFG gets away with it, I think, is because they do VSync a little differently and wait for it at a different place, which functions essentially like a sync. But I'm just guessing, here.

Since HMart reported performance drops with the current explicit glBufferData call, I've already switched to a triple-buffered approach with sync, just haven't pushed yet.
Downside is that the required glFenceSync is an OpenGL 3.2 feature, or an arb extension. But I think any reasonable driver today should probably support it.

renderer/BufferObject.cpp
stgatilov: I think it 1) copies a bit faster 2) makes sure that write destination does NOT get into CPU cache.
Not sure if the second point makes code faster. I recall that some special types of memory (like "pinned" or something like that) are better accessed with non-caching writes.

renderer/BufferObject.cpp
stgatilov: ARB_sync depends on OpenGL 3.1, so it is safe to assume that anything without OpenGL 3.1 will stop working.
I wonder if it would be possible to make it optional. Maybe we can insert some stupid glFinish if ARC_sync is missing? If it is possible and won't be horrendously slow, and it will allow for older OpenGL implementations to work at least somehow.

fholger: We can do that. But even BFG required OpenGL 3 core profile. I wonder if there are really any GPUs out there that are not compatible with OpenGL 3.1 and are able to play TDM otherwise?

stgatilov: Well, I know one case: Chromium-based driver in VirtualBox's Linux VM.
I have used it a lot for reviving linux build and even fixing bugs/crashes in it, despite the fact that it is barely playable. If this stops working completely, then TDM will be left with zero Linux-side developers again =(
I installed real Linux some years ago, and my experience shows that I never used it. In case of VM, I can even use Visual Studio for code browsing, combined with gdb-based debuggers =)
I understand that this stupid case does not justify anything...

fholger: Hm, fair point. But if performance doesn't matter, couldn't it potentially work with a Mesa software driver? I think that one is fairly advanced feature-wise.
In any case, I implemented glFenceSync for now without fallback, just to see if it works as intended. I'll keep this in mind once I've addressed any other issues.

renderer/BufferObject.cpp
fholger: I changed the flag.

renderer/BufferObject.cpp
fholger: Finally found it. BFG also uses glFenceSync, but they do it immediately after SwapBuffers. This is why they get away with 2 buffers without explicit orphaning. They do use glFinish as a fallback.
I think triple-buffering is a cleaner solution, since OpenGL can definitely queue more than one frame, so syncing on SwapBuffers could still cause stalls.

fholger: I added the fallback to glFinish.

renderer/VertexCache.cpp
```diff
@@ -18,207 +18,175 @@ Project: The Dark Mod (http://www.thedarkmod.com/)
...
 #include "tr_local.h"
 
-static const int	FRAME_MEMORY_BYTES = 0x200000;
-static const int	EXPAND_HEADERS = 32;
-
-idCVar idVertexCache::r_showVertexCache( "r_showVertexCache", "0", CVAR_INTEGER | CVAR_RENDERER, "" );
-idCVar idVertexCache::r_useMapBufferRange( "r_useMapBufferRange", "0", CVAR_BOOL | CVAR_RENDERER | CVAR_ARCHIVE, "use ARB_map_buffer_range for optimization" );
-idCVar r_queueVertexUploads( "r_queueVertexUploads", "0", CVAR_BOOL | CVAR_RENDERER | CVAR_ARCHIVE, "queue some vertex uploads from frontend to backend" );
+idCVar idVertexCache::r_showVertexCache( "r_showVertexCache", "0", CVAR_INTEGER | CVAR_RENDERER, "Show VertexCache usage statistics" );
+idCVar idVertexCache::r_staticVertexMemory( "r_staticVertexMemory", "40960", CVAR_INTEGER | CVAR_RENDERER, "Amount of static vertex memory, in kB (max 65535)" );
```
fholger, line 27: I made the cache memory sizes adjustable so that we don't have to set them to an initial high (save) value. But if someone does hit the limit, they at least have the possibility to increase it.

PR #6, opened 2018-06-20
FrameBuffer refactor

fholger:


renderer/FrameBuffer.h
```diff
@@ -46,3 +46,58 @@ void FB_BindShadowTexture();
...
+	FrameBuffer *primary;
+	FrameBuffer *resolve;
+	FrameBuffer *lightgem;
+	FrameBuffer *backBuffer;
+	FrameBuffer *finalOutput;
+
+	FrameBuffer *currentDraw;
+	FrameBuffer *currentRead;
+
+	std::vector<FrameBuffer*> allFrameBuffers;
```
stgatilov: I suggest using idLib instead of STL in renderer.
Given that renderer is mostly written in pure C, idLib fits much better: it is very simple an lightweight.

renderer/FrameBuffer.cpp
```diff
@@ -464,3 +464,221 @@ void FB_TogglePrimary( bool on ) {
...
+void FrameBuffer::CreateColorBuffer() {
+	if( colorBuffer == 0 ) {
+		qglGenRenderbuffers( 1, &colorBuffer );
+		if( colorBuffer == 0 ) {
+			common->FatalError( "Failed to allocate color render buffer" );
+		}
+	}
+	qglBindRenderbuffer( GL_RENDERBUFFER, colorBuffer );
+	if( msaa > 1 ) {
+		qglRenderbufferStorageMultisample( GL_RENDERBUFFER, msaa, GL_RGBA, width, height );
```
stgatilov, line 682: Since we are eager to support EXT_framebuffer extensions for 2.07, it is probably worth checking if this function pointer is not NULL, with fallback to creating non-multisampled renderbuffer if function is missing.
This is one tiny fix to line 547 =)

EDIT: or maybe it would be better to do this check in `FrameBuffer::Create` and change msaa right there.

renderer/FrameBuffer.cpp
```diff
@@ -464,3 +464,221 @@ void FB_TogglePrimary( bool on ) {
...
+	Bind();
+	qglFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_RENDERBUFFER, depthStencilBuffer );
+}
+
+void FrameBuffer::AddColorTexture( idImage *colorTexture ) {
+	if( msaa > 1 ) {
+		common->FatalError( "Rendering to texture for a framebuffer with AA enabled is not currently supported." );
+	}
+	if( colorBuffer != 0 ) {
+		common->Warning( "Setting framebuffer color attachment to texture, but a color render buffer was previously created." );
```
stgatilov: Doesn't it make sense to put assert here?
I think this case is no less serious than the one just above which throws FatalError. It shows a clear logical error somewhere in renderer, and it is quite unlikely that things will work well afterwards. In comparison, ignoring `msaa` can have no cnosequences except for lack of antialiasing.

renderer/FrameBuffer.cpp
```diff
@@ -464,3 +464,221 @@ void FB_TogglePrimary( bool on ) {
...
+
+void FrameBuffer::AddDepthStencilTextures( idImage *depthTexture, idImage *stencilTexture ) {
+	if( msaa > 1 ) {
+		common->FatalError( "Rendering to texture for a framebuffer with AA enabled is not currently supported." );
+	}
+	if( depthStencilBuffer != 0 ) {
+		common->Warning( "Setting framebuffer depth/stencil attachment to texture, but a render buffer was previously created." );
+	}
+	Bind();
+	qglFramebufferTexture2D( GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, depthTexture->texnum, 0 );
```
stgatilov, line 741: I think attached textures should be saved in Framebuffer somehow, not only in GL object. It makes it easier to debug, and add some possibilities.
For instance, you forbid attaching a color texture to FBO with color renderbuffer already created. But you do not forbid creating color renderbuffer when color texture is already attached. Is it intentional? It is hard to think that engine is operating properly when it first attaches a texture to FBO and then switches to renderbuffer for the rest of time...

renderer/FrameBuffer.cpp
```diff
@@ -464,3 +464,221 @@ void FB_TogglePrimary( bool on ) {
...
+		qglRenderbufferStorageMultisample( GL_RENDERBUFFER, msaa, GL_DEPTH24_STENCIL8, width, height );
+	} else {
+		qglRenderbufferStorage( GL_RENDERBUFFER, GL_DEPTH24_STENCIL8, width, height );
+	}
+
+	Bind();
+	qglFramebufferRenderbuffer( GL_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_RENDERBUFFER, depthStencilBuffer );
+}
+
+void FrameBuffer::AddColorTexture( idImage *colorTexture ) {
```
stgatilov, line 714: Is there any way to ensure that images does not die before FBO is deleted?
Renderbuffers are owned by FBO, so they won't die before FBO dies.
But it might happen so that idImage (or its gl texture object) is deleted before FBO is, which can cause serious issues.

stgatilov (commented):
Looks like a great idea in general!
Pretty much everything in OpenGL needs wrapping for convenient and safe usage =)

renderer/FrameBuffer.h
fholger: On the other hand, the STL has come a long way since id created idLib, and I believe I've seen Carmack mention that he was not overly happy with idLib. Modern vector implementations probably make use of C++1x features like move etc, giving better performance.

Of course, for this use case it's pretty much irrelevant, and I don't really have a strong preference.  I'll happily switch to idLib::List.

renderer/FrameBuffer.cpp
fholger: Yeah, the fallbacks are still TODO; there's also an ARB extension covering framebuffers including MSAA. My hope is that the function signatures are identical and we can implement most of the fallbacks during loading of the function pointers, but will have to see.

renderer/FrameBuffer.cpp
fholger: Well, it's a bit of a fuzzy line. Technically, reassigning the color attachment to a texture after the fact will still leave the FBO functional, so not really a fatal error, but it's probably wrong. So I went with a warning here. Possibly asserts would be the better alternative.

The MSAA case is a bit trickier. If you mix an AA renderbuffer with a non-AA texture, you will end up with an invalid framebuffer, so that's closer to a real error. Of course, if you assign textures everywhere, it would end up working. But yeah, probably should make the error handling consistent.

renderer/FrameBuffer.cpp
fholger: Fair point. I went for the simplest initial implementation, and since we don't make use of the textures after binding them, I omitted them.

renderer/FrameBuffer.cpp
fholger: It would render the FBO invalid, definitely. However, I don't see a clear way to prevent that. You would need a different ownership model for idImage to stop someone from deleting it before the FBO is released.
We could perhaps add a kind of marker to a texture, but that doesn't strike me as particularly elegant.

renderer/FrameBuffer.cpp
stgatilov: See svn rev 7501.
The functions and constants are same, although some completeness conditions may differ slightly.

So checking the function pointer is now a sane way of checking whether MS+FBO is supported by any means.

renderer/FrameBuffer.cpp
stgatilov: Isn't there a global manager for all `idImage`-s ? I guess `idImageManager *globalImages` works like this. All images are allocated via `idImageManager::AllocImage` method and deallocated via `idImage::PurgeImage` method.

In this case it is possible to check if an image is attached to FBO when it is purged, and raise an error in this case. Of course, this is not exactly trivial, you have to read through idImage and idImageManager code and make sure you understand how it works.

PR #7, opened 2018-07-30
Enable GL debug groups and profiling via timer queries

fholger:


renderer/Profiling.cpp
```diff
@@ -0,0 +1,181 @@
...
+ * be accumulated to a single total time in the end.
+ */
+class GlProfiler {
+public:
+	GlProfiler() : profilingActive(false), frameMarker(0) {}
+
+	void BeginFrame() {
+		sectionStack.clear();
+		frame[ frameMarker ] = section();
+		frame[ frameMarker ].name = "Frame";
```
stgatilov: Maybe add frame index here (I guess we have some numbering of frames)...

renderer/Profiling.cpp
```diff
@@ -0,0 +1,181 @@
...
+		idStr name;
+		int count;
+		double totalCpuTimeMillis;
+		double totalGpuTimeMillis;
+		std::vector<section> children;
+		std::vector<query> queries;
+	} section;
+
+	section * CollectTimings() {
+		AccumulateTotalTimes( frame[ frameMarker ] );
```
stgatilov: Do I understand right that here you request timings of the **current** frame?
If yes, than how can it display CPU timings correctly?
What if this function is called when profiler is active?

It seems that the most non-intrusive policy is to display results of the previous or one-before-previous frame, while storing more than two consecutive frames.
Fetching results of the just finished frame is not a very good idea, because if the rendering of that frame is not yet finished, then quering gl timers would block CPU (If I understand correctly how it works).

renderer/Profiling.cpp
```diff
@@ -0,0 +1,181 @@
...
+	typedef struct section {
+		idStr name;
+		int count;
+		double totalCpuTimeMillis;
+		double totalGpuTimeMillis;
+		std::vector<section> children;
+		std::vector<query> queries;
+	} section;
+
+	section * CollectTimings() {
```
stgatilov: This function should be guarded against double call, I guess...

renderer/Profiling.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef __PROFILING_H__
+#define __PROFILING_H__
+
+#include "qgl.h"
+#include "tr_local.h"
+
+extern idCVar r_useDebugGroups;
+extern idCVar r_glProfiling;
+
+void EnterProfilingSection( const char* section );
```
stgatilov: I'll probably grumble about several things here.

The first one is naming. I know that renderer is mostly C-style code, with separate functions declared in h-files and defined somewhere in cpp. But anyway it's better to start all header-available function names with common prefix. If it is "Profiling", then better rename functions to `ProfilingLeaveSection` and `ProfilingDisplayInfo`.
And when you change all names to have similar prefix, it will become apparent that it makes more sense to group the functions into a namespace at least (or even class) =) Of course, if we follow the C-style, then the only grouping that we can use is having a common prefix in function names =)

The second one is about exposing `EnterProfilingSection` and `LeaveProfilingSection`. What is the intended usage of it? Right now `GlProfileScope` is guarded against absence of GL functionality (probably not needed) and also checks for cvars. The functions in question do not check for anything.

sys/scons/SConscript.core
```diff
@@ -52,6 +52,7 @@ renderer_string = ' \
 	Model_md5.cpp \
 	Model_prt.cpp \
 	Model_sprite.cpp \
+    Profiling.cpp \
```
stgatilov: tab?...

stgatilov (commented):
One more question:
What exactly are the changes to `renderer/glext.h` ?

renderer/Profiling.cpp
fholger: You do understand correctly, but the code already double-buffers. The sequence of events is always BeginFrame, EndFrame, CollectTimings. EndFrame finishes the current frame, then switches the frameMarker, so CollectTimings will actually access the timings from the previous frame (which is then recycled by the next BeginFrame). Yes, I admit that's confusing and I don't particularly like it. I'm open to ideas on how to structure this in a less confusing manner :)

sys/scons/SConscript.core
fholger: Whoops. Used the wrong editor :D

renderer/Profiling.h
fholger: Let's begin with the second part. The reason they are exposed is because I wanted to have the GlProfileScope constructor and destructor in the header, so that the compiler could inline the cvar guards and keep impact on the runtime minimal if they are disabled. For that, the functions need to be exposed in the header, too.
But I'll readily admit that I probably fell into the "don't optimize prematurely" trap here. Functionally, there is no reason to expose the functions nor should they ever be called directly.

As for the naming: yeah, namespace would be the ideal solution, but it kind of breaks the Doom3 style. I'm fine with putting Profiling as a prefix, though.

renderer/Profiling.cpp
fholger: We do, but I'm not sure how readily it is accessible from this part of the code. If it's not too invasive, I'll add it.

fholger: Btw, I can't seem to answer to your glext question directly, so I'll put it here: I had to update the glext.h with a newer version directly from the Khronos group. The reason is that our current version is at an age where it does not contain any GL4 stuff, so the glPushDebugGroup and glPopDebugGroup functions are missing. The update shouldn't break anything, it's just updated extension definitions, basically.

stgatilov: Well, unless previous version had local modification...
Oh God! How could I even think of it =)

renderer/Profiling.cpp
stgatilov: At least add here assert that profiling is not active.
This will ensure that the frame you are processing here is already finished (from CPU prespective).

renderer/Profiling.cpp
fholger: Not very likely, and if so, all the more reason to clean it up :D
But again, the glext.h only contains type information about all available extension methods, it does not actually provide any implementations. So the very fact that TDM still compiles with the new version almost certainly guarantees that the update didn't break anything.

PR #9, opened 2019-02-16
WIP: Shaders

fholger:


renderer/GLSLProgram.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+void GLSLProgram::Deactivate() {
+	qglUseProgram( 0 );
+	currentProgram = 0;
+}
+
+void GLSLProgram::AddUniformAlias( int alias, const char *uniformName ) {
+	int location = qglGetUniformLocation( program, uniformName );
+	if( location == -1 ) {
+		common->Warning( "Did not find active uniform: %s\n", uniformName );
```
stgatilov, line 62: This should **not** be a warning.

It is very frequent that uniform is declared in the program but not used. In such cases most OpenGL implementations remove uniform entirely, and this warning will be raised, despite the fact that uniform is declared. This will cause a lot of confusion and noise.

I'm afraid we have to silently map this `alias` into `-1` location.

renderer/GLSLProgram.h
```diff
@@ -0,0 +1,109 @@
...
+#ifndef __GLSL_PROGRAM_H__
+#define __GLSL_PROGRAM_H__
+
+class GLSLProgram {
+public:
+	static GLSLProgram * Load( const char *vertexSourceFile, const char *fragmentSourceFile, const char *geometrySourceFile = nullptr );
+	static GLSLProgram * Load( const idDict &defines, const char *vertexSourceFile, const char *fragmentSourceFile, const char *geometrySourceFile = nullptr );
+
+	GLSLProgram( GLuint program );
+	~GLSLProgram();
```
stgatilov, line 24: Since GLSLProgram owns program object, it makes sense to forbid copying --- just in case someone tries =)

renderer/GLSLProgram.cpp
```diff
@@ -0,0 +1,334 @@
...
+	GLuint shader = qglCreateShader( shaderType );
+	GLint length = source.size();
+	const char *sourcePtr = source.c_str();
+	qglShaderSource( shader, 1, &sourcePtr, &length );
+	qglCompileShader( shader );
+
+	// check if compilation was successful
+	GLint result;
+	qglGetShaderiv( shader, GL_COMPILE_STATUS, &result );
+	if( result == GL_FALSE ) {
```
stgatilov, line 305: I wish warnings were displayed even when compilation is successful.

It is not viable right now (because there would be too many warnings), but I hope it will be done somewhere in the bright future =)

renderer/GLSLProgram.h
```diff
@@ -0,0 +1,109 @@
...
+
+#ifndef __GLSL_PROGRAM_H__
+#define __GLSL_PROGRAM_H__
+
+class GLSLProgram {
+public:
+	static GLSLProgram * Load( const char *vertexSourceFile, const char *fragmentSourceFile, const char *geometrySourceFile = nullptr );
+	static GLSLProgram * Load( const idDict &defines, const char *vertexSourceFile, const char *fragmentSourceFile, const char *geometrySourceFile = nullptr );
+
+	GLSLProgram( GLuint program );
```
stgatilov: `explicit` to avoid stupid trouble? =)

renderer/GLSLProgram.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+	GLuint shader = CompileShader( shaderType, sourceFile, defines );
+	if( shader != 0) {
+		qglAttachShader( program, shader );
+		// won't actually be deleted until the program it's attached to is deleted
+		qglDeleteShader( shader );
+	}
+}
+
+std::string ReadFile( const char *sourceFile ) {
```
stgatilov: Maybe static/namespace to make symbol private...

renderer/GLSLProgram.cpp
```diff
@@ -0,0 +1,334 @@
...
+			// unfortunately, #line does not take an actual filename, but only an integral reference to a file :(
+			unsigned int currentLine = std::count( source.cbegin(), match[ 0 ].first, '\n' ) + 1 - totalIncludedLines;
+			std::string includeBeginMarker = "#line 0 " + std::to_string( nextFileNo ) + '\n';
+			std::string includeEndMarker = "\n#line " + std::to_string( currentLine ) + ' ' + std::to_string( currentFileNo );
+			totalIncludedLines += std::count( includeContents.begin(), includeContents.end(), '\n' ) + 2;
+
+			// replace include statement with content of included file
+			source.replace( match[ 0 ].first, match[ 0 ].second, includeBeginMarker + includeContents + includeEndMarker );
+		} else {
+			source.replace( match[ 0 ].first, match[ 0 ].second, "// already included " + fileToInclude );
```
stgatilov: BTW, non-guarded includes allow [a useful trick](https://stackoverflow.com/a/17670232/556899) in C-like language. Let's hope no one will ever want this in GLSL shaders =)))

renderer/GLSLProgram.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+	return shader;
+}
+
+globalPrograms_t globalPrograms { nullptr };
+
+void GLSL_InitPrograms() {
+	idDict stencilShadowDefines;
+	stencilShadowDefines.Set( "SHADOW_TYPE", "1" );
+	globalPrograms.stencilInteractionShader = GLSLProgram::Load( stencilShadowDefines, "glprogs/interaction.vs", "glprogs/interaction.fs" );
```
stgatilov: How does include paths work?

Given that _glprogs_ is standard directory for all shaders, is it in include path?

renderer/GLSLProgram.cpp
fholger: That depends strongly on the use case. My expectation is that at least for the core shaders we only pass in information that the shader actually needs. So if the uniform isn't actually used in the shader, that could hint at a misspelling or other mistakes, and so the warning could be quite helpful.

If we treat shaders more generically than that, then yes, the warning might be counter-productive.

renderer/GLSLProgram.h
fholger: Yes, absolutely.

renderer/GLSLProgram.cpp
fholger: Do GLSL compilers actually produce useful warnings? We could put output behind a cvar, so it's off by default, but accessible for shader code writing and debugging.

renderer/GLSLProgram.h
fholger: Yes. Perhaps even private and friend the GLSLProgramLoader class, because at least in theory the constructor isn't meant to be called directly?

renderer/GLSLProgram.cpp
fholger: Well, that's the point where we'd have to actually look at a full preprocessor solution ;)

renderer/GLSLProgram.cpp
fholger: The path is relative to the game directory. So if a shader lies in glprogs, you'd have to include it via glprogs/shader_file. I did not implement any search path logic on top of that and think it might be overkill. But if we want to restrict shaders to the glprogs directory, we could also make that the base path for the include directive.

PR #12, opened 2019-10-03
CMake build system

fholger:


sys/cmake/SvnVersion.cmake
```diff
@@ -0,0 +1,17 @@
+macro(get_svn_revision OutVar)
+    message("Determining SVN revision")
+    find_program(SVNVERSION_PROGRAM name svnversion)
```
stgatilov: It should be `NAMES` instead of `name` here.
Otherwise it gives me "subversion not found" error.

I'm quite puzzled that the NAMES keyword is case-sensitive.

sys/cmake/SvnVersion.cmake
```diff
@@ -0,0 +1,17 @@
...
+    if (SVNVERSION_PROGRAM)
+        execute_process(COMMAND "${SVNVERSION_PROGRAM}" "-c"
+                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+                RESULT_VARIABLE SVNVERSION_SUCCESS
+                OUTPUT_VARIABLE ${OutVar})
+        if (NOT SVNVERSION_SUCCESS)
```
stgatilov: Had to replace it with:

        if (NOT SVNVERSION_EXITCODE EQUAL 0)

Otherwise I got "svnversion failed". I have also renamed the variable, because it is the exit code, which is actually "false" on success.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+target_sources(TheDarkMod PRIVATE ${TDM_SOURCE_FILES})
+
+if (MSVC)
+	ucm_add_dirs("sys/msvc/natvis" TO TDM_NATVIS_FILES RECURSIVE ADDITIONAL_EXT "natvis")
+	target_sources(TheDarkMod INTERFACE ${TDM_NATVIS_FILES})
+endif()
+
+# configure precompiled header
+# note TypeInfo.cpp must be excluded from precompiled header because it uses some dirty hacks
+add_precompiled_header(TheDarkMod "idlib/precompiled.h" "idlib/precompiled.cpp" "precompiled.h"
```
stgatilov, line 100: Does not work for me. Compiler complains that some language constructs require at least C++11.

Note that you have set C++14 above, and that my compiler (GCC 5.4.0) seems to support it.
Maybe the precompiler header macro does not propagate C++ standard setting properly...

stgatilov: Same with this line commented out.
Somehow, C++14 is dropped from compile flags.

stgatilov (commented):
This is preliminary review.
I got stuck with missing C++11/14 compile options on my machine.

CMakeLists.txt
stgatilov: Adding `add_definitions(-std=c++14)` fixes the problem, as long as I keep PCH disabled. If I enable PCH, it does not compile complaining about the lack of c++11. But the hell is wrong with me?...

fholger: It's probably either the CMake or GCC version which behaves differently than the ones I've tested with. Just for reference, what version of CMake are you using? If it's not 3.15, could you (easily) test with 3.15 just to rule that out?

stgatilov: I'm using GCC 5.4.0, which is stock on Ubuntu 16.04, and CMake 3.15.5, which I had to build myself.

sys/cmake/SvnVersion.cmake
```diff
@@ -0,0 +1,17 @@
+macro(get_svn_revision OutVar)
+    message("Determining SVN revision")
+    find_program(SVNVERSION_PROGRAM name svnversion)
+    if (SVNVERSION_PROGRAM)
+        execute_process(COMMAND "${SVNVERSION_PROGRAM}" "-c"
```
stgatilov: Another problem is that svnversion prints EOL at the end of its output, which somehow gets into the variable. As the result, the last double-quote is put on a separate line in svnversion.h. It has to be removed somehow.

stgatilov: Adding OUTPUT_STRIP_TRAILING_WHITESPACE parameter to this command removes the damn EOL for me.

CMakeLists.txt
stgatilov: I guess I have found the problem: you perform add_executable too early. The cxx_standard is specified later, and it seems that it does not affect TheDarkMod executable. To be honest, it would be rather strange if setting a global variable affected the projects defined later... although this is cmake =) Could you please clarify which order of statements should be used to make everything work?

I have tried to put add_executable immediately before you add source files to it, and in that case C++14 got assigned normally (without hacky add_definitions). However, as soon as I uncomment precompiled header, it stops working again --- PCH also does not catch the c++14 flag.

fholger: Thanks for your analysis. I guess newer GCC versions might derive C++14 from the project somehow, or perhaps it's just the default.
I'll try and hack the C++14 flags into the precomnpiled header options somehow.

sys/cmake/SvnVersion.cmake
fholger: Sorry for the svnversion troubles. I didn't really get to test this part due to my git-svn setup. So thanks for trouble-shooting it, I'll get the necessary changes in :)

fholger: Should be working now, hopefully.

CMakeLists.txt
fholger: I've added `-std=c++14` explicitly to the compile options in general and for the precompiled header. My assumption is that this setting is default in later versions, but not yours, and for some reason CMake doesn't propagate it properly. But unfortunately, I can't test it because the oldest GCC version I have access to right now is 6, and that just works. Would appreciate it if you gave it another quick spin. If that doesn't fix it, we may have to disable PCH for older GCC versions...

stgatilov: Yes, it seems that GCC 6 has enabled at least C++11 by default.
You can temporarily add `constexpr if` into some common header and test it with C++17 standard =)

stgatilov:
With the new changes, std=c++14 is enabled everywhere. Although setting standard in three places is dirty... probably we can try to clean it up later.

Right now I get the following error with unmodified state:
```
[ 17%] Building CXX object CMakeFiles/TheDarkMod.dir/game/Mover.cpp.o
/tmp/ccpToZJw.s: Assembler messages:
/tmp/ccpToZJw.s:5516: Error: symbol `_ZN7idMoverUt_C2Ev' is already defined
CMakeFiles/TheDarkMod.dir/build.make:1192: recipe for target 'CMakeFiles/TheDarkMod.dir/game/Mover.cpp.o' failed
make[2]: *** [CMakeFiles/TheDarkMod.dir/game/Mover.cpp.o] Error 1
```

With precompiled header disabled, it builds fine.

CMakeLists.txt
fholger: I've excluded Mover.cpp from the precompiled header targets. Could you check if that solves it? I have no idea what's going on, if that doesn't fix it, we may have to disable PCH for GCC below 6.

stgatilov: I have fixed the problem in svn rev 8424. You can revert this "exclusion" back.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
+cmake_minimum_required(VERSION 3.12)
```
stgatilov, line 1: Let's downgrade to version 3.5.

This version is the highest available on stock Ubuntu 16.04 LTS. And we don't use any newer features, I think.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+include(ucm)
+include(PrecompiledHeader)
+include(SvnVersion)
+
+project(TheDarkMod VERSION 2.08)
+add_executable(TheDarkMod)
+
+cmake_dependent_option(WITH_TOOLS "Compile with Tools" ON "MSVC" OFF)
+set(GAME_DIR "${PROJECT_SOURCE_DIR}/../darkmod" CACHE STRING "The directory where the game assets are stored")
+option(COPY_EXE "Copy executable to GAME_DIR after build" ON)
```
stgatilov, line 25: There is one important question about debug information. I guess I'll better write it on forums.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+get_svn_revision(SVNVERSION)
+configure_file("idlib/svnversion.cmake.h" "${PROJECT_SOURCE_DIR}/idlib/svnversion.h" @ONLY)
+
+# gather list of source files
+ucm_add_dirs(
+		"cm"
```
stgatilov, line 34: Sorry to post such stupid comments, but I really dislike indentation in this file. It is weird and inconsistent.

In my opinion, `*.cmake` and `CMakeLists.txt` should be added to `.editorconfig`, just like scons files are currently present there. After that, stick to one-tab indentation everywhere.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+
+if (UNIX)
+	ucm_add_dirs("sys/linux" "sys/posix" TO TDM_SOURCE_FILES RECURSIVE)
+    ucm_remove_directories("idlib/sys/win32" FROM TDM_SOURCE_FILES)
+	ucm_remove_files(
+			"renderer/glad_wgl.c"
+			"sys/linux/dedicated.cpp"
+			"sys/linux/test_scheduler.c"
+			"sys/posix/platform_osx.cpp"
+			FROM TDM_SOURCE_FILES)
```
stgatilov: I would say: this closing bracket should be on separate line. That would make nested structure more clear, and will be consistent with the rest of the file.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+			)
+	add_compile_options("/MP /Gm-")
+	if (CMAKE_SIZEOF_VOID_P EQUAL 8)
+		set(PACKAGE_PLATFORM "win64_s_vc15_rel_mt")
+		set(PACKAGE_PLATFORM_DEBUG "win64_s_vc15_dbg_mtd")
+		set_target_properties(TheDarkMod PROPERTIES OUTPUT_NAME "TheDarkModx64")
+    else ()
+		set(PACKAGE_PLATFORM "win32_s_vc15_rel_mt")
+		set(PACKAGE_PLATFORM_DEBUG "win32_s_vc15_dbg_mtd")
+	endif()
```
stgatilov: Probably set OUTPUT_NAME here so that people don't guess what is default...

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+			-Wno-class-memaccess            # memcpy into custom structs
+	)
+	# configure other compile options
+	add_compile_options(
+			-pipe
+			-fno-strict-aliasing
+			-fmessage-length=0
+			-fvisibility=hidden
+			-fno-unsafe-math-optimizations
+	)
```
stgatilov, line 130: One key is missing: 

	# use old ABI for std::string and std::list (which is not fully compliant with C++11 standard)
	# this allows to run TDM binary on OSes which have old glibcxx (e.g. Ubuntu 14.04 from years 2014-2016)
	BASECPPFLAGS.append( '-D_GLIBCXX_USE_CXX11_ABI=0' )

I'm pretty sure someone with old-as-hell OS will come and complain that TDM does not start =) This is how this key was set in SCons. Better add it here.

sys/cmake/gcc_32bit.cmake
```diff
@@ -0,0 +1,11 @@
+# To compile for 32bit platform on Linux, call CMake like this:
+# cmake path/to/source -DCMAKE_TOOLCHAIN_FILE=path/to/source/sys/cmake/gcc_32bit.cmake
+
+string(FIND "${CMAKE_CXX_FLAGS}" "-m32 -msse3" M32_POS_CXX)
```
stgatilov: Some questions about SIMD option:

1. Why enable SSE3? Only SSE2 is enabled by default on x64, and the same is true for our current Linux builds.
2. Why setting this key in platform settings? It seems to be better to set it along with other Linux compile keys. Then this toolchain file would be optional on 32-bit OS, which can be reflected in the comment (which now says that this toolchain file is required for 32-bit build regardless of OS).

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+			NO_WARN_MBCS_MFC_DEPRECATION
+			_MBCS
+			)
+	add_compile_options("/MP /Gm-")
+	if (CMAKE_SIZEOF_VOID_P EQUAL 8)
+		set(PACKAGE_PLATFORM "win64_s_vc15_rel_mt")
+		set(PACKAGE_PLATFORM_DEBUG "win64_s_vc15_dbg_mtd")
+		set_target_properties(TheDarkMod PROPERTIES OUTPUT_NAME "TheDarkModx64")
+    else ()
+		set(PACKAGE_PLATFORM "win32_s_vc15_rel_mt")
```
stgatilov, line 165: While there are a lot of build customizations in VC projects, one option must be enabled, and it is `/arch:SSE2`. Even though it is stated as enabled by default in Microsoft docs, I'm not certain that it is default in CMake-generated VC project.
The idea of having TDM build with x87 math is scary...

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
+cmake_minimum_required(VERSION 3.12)
+cmake_policy(SET CMP0076 NEW)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/sys/cmake")
+include(CMakeDependentOption)
+include(ucm)
+include(PrecompiledHeader)
+include(SvnVersion)
+
+project(TheDarkMod VERSION 2.08)
```
stgatilov: I wonder if this has any meaning.
It is probably easier to not set any version, since otherwise there is always risk of forgetting to update the version somewhere =)

stgatilov (approved):

CMakeLists.txt
fholger: Unfortunately, we do. `add_compile_options` and the particular syntax of `add_executable` used, for example, don't work with lower versions. This could potentially be ported, but it is more effort than simply lowering the version number and loses some clarity in the process.

CMakeLists.txt
fholger: Yeah, probably. This could make sense if CMake were the only build system and could be the actual source of truth for the version, but otherwise it's just going to get out of sync :)

CMakeLists.txt
stgatilov: Downloaded CMake 3.5.1, and indeed you are right.

On the other hand, Ubuntu 18.04 has only CMake 3.10. Given that a few years ago we had players with 14.04, I'm afraid we will have questions about how to install CMake. This is not a stopper for merging the changes of course, but I wonder how do you feel in general about backporting it to 3.5 in future.

fholger: I'm a bit conflicted. On the one hand, I do understand the argument and it would be nice if it worked with the distro-shipped versions. On the other hand, newer CMake versions do add useful new features, and having to restrict yourself to outdated dev tools is painful. I'm not very happy with the Debian/Ubuntu conservative update policy on dev tools, in general :)

The question for me is how many users will actually attempt to compile TDM who do not have the technical expertise to install CMake manually. I'm inclined to wait and see if there's an actual demand before investing time in a backport, because I think that time could be better spent at the moment.

fholger: BTW, CMake is available as a Snap package, and given that Ubuntu seems to push Snap packages in general, that may lessen the importance of the APT version?

stgatilov: Yes, snap is a good idea. I should try it out =)
Let's forget about this question then.

PR #17, opened 2020-04-15
New bloom stage

fholger:


engine.vcxproj
```diff
@@ -1001,6 +1002,7 @@
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Release|x64'">Create</PrecompiledHeader>
     </ClCompile>
     <ClCompile Include="renderer\AmbientOcclusionStage.cpp" />
+    <ClCompile Include="renderer\BloomStage.cpp" />
```
stgatilov, line 1005: Remainder: update Scons files too.

I'd like to maintain them for 2.08 release as a secondary option.

glprogs/ambientInteraction.fs
```diff
@@ -90,6 +90,7 @@ void main() {
 		light.rgb *= matDiffuse.rgb * light1;
 	}
 
+	light = max(light, vec4(0));  // avoid negative values, which with floating point render buffers can lead to NaN artefacts
```
stgatilov, line 93: Can't see a way how this could be negative, as long as all input colors and textures are nonnegative of course.

glprogs/bloom_blur.frag.glsl
```diff
@@ -0,0 +1,30 @@
...
+
+// (1, 0) or (0, 1)
+uniform vec2 u_axis;
+
+vec2 invImageSize = vec2(1) / vec2(textureSize(u_source, u_mipLevel));
+
+
+void main() {
+    vec2 fragCoord = gl_FragCoord.xy;
+	FragColor = textureLod(u_source, fragCoord * invImageSize, u_mipLevel) * weight[0];
```
stgatilov: indent?

glprogs/bloom_downsample.frag.glsl
```diff
@@ -0,0 +1,41 @@
...
+	return clamp(pow(brightness / u_brightnessThreshold, u_thresholdFalloff), 0, 1);
+}
+#endif
+
+vec4 sampleTexture(vec2 offset) {
+	return textureLod(u_sourceTexture, var_TexCoord + offset, u_sourceMipLevel);
+}
+
+void main() {
+	vec2 halfpixel = 2 * vec2(0.5, 0.5) / textureSize(u_sourceTexture, u_sourceMipLevel);
```
stgatilov: Is it still **half**-pixel?

If yes, maybe write a comment here?

glprogs/tonemap.fs
```diff
@@ -35,6 +37,10 @@ float mapColorComponent(float value) {
 
 void main() {
 	vec4 color = texture(u_texture, var_TexCoord.xy);
+    if (u_bloomWeight > 0) {
```
stgatilov: Well... I wanted to keep this shader free from any complicated stuff. But performance-wise, it makes sense to add bloom here instead of doing yet another pass.

renderer/BloomStage.cpp
```diff
@@ -0,0 +1,321 @@
...
+		qglTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAX_LEVEL, BloomStage::NUM_FILTER_STEPS - 1);
+		qglTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR_MIPMAP_NEAREST);
+		qglTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_NEAREST);
+		qglTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+		qglTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+		GL_SetDebugLabel(GL_TEXTURE, image->texnum, image->imgName);
+	}
+}
+
+BloomStage::BloomStage() : downsampleFBOs{ 0 }, upsampleFBOs{ 0 }, bloomDownSamplers( nullptr ), bloomUpSamplers( nullptr ),
```
stgatilov: Hmm... some member not initialized?

Maybe it would be easier to write initial values in class body, in C++11 style...

renderer/BloomStage.cpp
```diff
@@ -0,0 +1,321 @@
...
+		blurShader = programManager->LoadFromGenerator("bloom_blur", LoadBloomBlurShader);
+	}
+	upsampleShader = programManager->Find("bloom_upsample");
+	if (upsampleShader == nullptr) {
+		upsampleShader = programManager->LoadFromGenerator("bloom_upsample", LoadBloomUpsampleShader);
+	}
+}
+
+void BloomStage::Shutdown() {
+	if (downsampleFBOs[0] != 0) {
```
stgatilov, line 208: Shutdown looks incomplete.
What about other FBOs and samplers?

renderer/BloomStage.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+	uniforms->axis.Set( 1, 0 );
+	qglBindFramebuffer(GL_FRAMEBUFFER, upsampleFBOs[mipLevel]);
+	qglViewport(0, 0, bloomUpSamplers->uploadWidth >> mipLevel, bloomUpSamplers->uploadHeight >> mipLevel);
+	qglClear(GL_COLOR_BUFFER_BIT);
+	RB_DrawFullScreenQuad();
+
+	uniforms->axis.Set( 0, 1 );
+	qglBindFramebuffer(GL_FRAMEBUFFER, downsampleFBOs[mipLevel]);
+	bloomUpSamplers->Bind();
```
stgatilov, line 318: I must say that setting texture unit uniforms beforehand makes the code much harder to understand. Perhaps that is an optimization for something which is executed hundreds of time per frame, but here it is not necessary.

renderer/BloomStage.cpp
```diff
@@ -0,0 +1,321 @@
...
+	GL_PROFILE("BloomBlur")
+
+	int mipLevel = NUM_FILTER_STEPS - 1;
+	blurShader->Activate();
+
+	BloomBlurUniforms *uniforms = blurShader->GetUniformGroup<BloomBlurUniforms>();
+	uniforms->mipLevel.Set( mipLevel );
+
+	uniforms->axis.Set( 1, 0 );
+	qglBindFramebuffer(GL_FRAMEBUFFER, upsampleFBOs[mipLevel]);
```
stgatilov, line 310: `downsampleFBOs` and `upsampleFBOs` are swapped here?...

renderer/BloomStage.cpp
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	if (downsampleFBOs[0] == 0) {
+		Init();
+	}
+
+	qglClearColor(0, 0, 0, 0);
+	Downsample();
+	Blur();
+	Blur();
```
stgatilov: I wonder what is done by Blur filter. I seems missing from original presentation. And why is it called twice?
Does it play nicely for upsampling later?

glprogs/ambientInteraction.fs
fholger: One would think so, and yet it happened. I didn't actually observe truly negative numbers, but in some rare cases there were rounding errors when nSight would then show the texture content as (-0.000...) instead of just (0), and that was enough to produce NaN and visual artifacts in coming calculations

renderer/BloomStage.cpp
fholger: Samplers and FBOs are all there? Only shaders are missing, because the manager takes care of them... (Note that the qglDeleteFramebuffers call deletes multiple FBOs. I only set the first explicitly to 0 as a marker.)

renderer/BloomStage.cpp
fholger: I originally started with an approach that did blurring just by down- and upsampling, but I felt that the blur achieved by doing so was not strong enough. Blur performs an actual Gaussian Blur at the lowest downsampling level to get that smoother, wider blur. And basically it's done twice to increase the radius of the blur. I'll add some comments to explain the approach.

renderer/BloomStage.cpp
fholger: The approach is basically:

- downsample to lowest mip in the downsamplers
- apply Gaussian blur to that lowest mip. This is a two-phase blur (vert/horz) and needs one intermediary texture to store the results of the first blur pass. I used the upsampling texture at lowest mip for that, because it seemed easier than creating yet another separate texture and framebuffer. Final result is stored back in the lowest mip of the downsamplers
- upsample, first from downsamplers[lowest_mip] to upsamplers[lowest_mip-1], then from upsamplers[mip+1] to upsamplers[mip]

renderer/BloomStage.cpp
fholger: Yes, fair enough. Will change :)

engine.vcxproj
fholger: Done.

glprogs/bloom_downsample.frag.glsl
fholger: From a certain point of view, it is. But you're right, it may be more confusing than helpful, so I changed it to a more generic `offset` and added a short comment.

glprogs/tonemap.fs
fholger: Yeah, I thought about it, but it seemed the most logical place to do it. On the upside, I think a texture lookup and color add is not too complicated :)

PR #18, opened 2020-05-01
C++14 check for CMake + fix for ThirdParty/devil + sysconf() in platform_linux.cpp

r-a-sattarov:
- Added check of compiler for support C++14 standard in CMake
- Fix in ThirdParty/devil for MCST lcc compiler
- Added alternative method to get s_numPhysicalCPUCores value (usable for non-x86 CPU)


fholger:
These changes have been applied by stgatilov to the SVN master.

r-a-sattarov:
Yes, I see this changes in this commit - https://github.com/fholger/thedarkmod/commit/7c9081d1837e6bf2df349bac7c2f3fa1cb39fc70. Many thanks! :+1:

PR #19, opened 2020-05-09
Fbo management

fholger:


renderer/FrameBuffer.cpp
```diff
@@ -21,765 +21,311 @@ Project: The Dark Mod (http://www.thedarkmod.com/)
...
-	GL_CheckErrors();
-
-	GL_Viewport( 0, 0, glConfig.vidWidth, glConfig.vidHeight );
-	GL_Scissor( 0, 0, glConfig.vidWidth, glConfig.vidHeight );
-
-	if ( r_multiSamples.GetInteger() > 1 ) {
-		FB_ResolveMultisampling( GL_COLOR_BUFFER_BIT );
-		qglBindFramebuffer( GL_READ_FRAMEBUFFER, fboResolve );
+void FrameBuffer::AddRenderTexture( idImage *texture, GLenum attachment, int mipLevel ) {
+	if (texture->uploadWidth >> mipLevel != width || texture->uploadHeight >> mipLevel != height) {
```
stgatilov: Bit operations without parentheses make me nervous =)

renderer/FrameBuffer.cpp
```diff
@@ -21,765 +21,311 @@ Project: The Dark Mod (http://www.thedarkmod.com/)
...
-	qglBindFramebuffer( GL_FRAMEBUFFER, 0 );
+renderCrop_t ShadowAtlasPages[42];
+idCVar r_fboResolution( "r_fboResolution", "1", CVAR_RENDERER | CVAR_FLOAT | CVAR_ARCHIVE, "internal rendering resolution factor" );
 
-	if ( r_frontBuffer.GetBool() ) {
-		qglFinish();
-	}
-	GL_CheckErrors();
+#if defined(_MSC_VER) && _MSC_VER >= 1800 && !defined(DEBUG)
+//#pragma optimize("t", off) // duzenko: used in release to enforce breakpoints in inlineable code. Please do not remove
```
stgatilov: Maybe ask Duzenko once again and remove this dirty thing?

renderer/FrameBuffer.cpp
```diff
@@ -789,16 +335,7 @@ if you absolutely must check for anything out of the ordinary, then do it here.
...
-	if ( w <= 0 || h <= 0 )
-		return;
-	if ( primaryOn ) {
-		x *= r_fboResolution.GetFloat();
-		y *= r_fboResolution.GetFloat();
-		w *= r_fboResolution.GetFloat();
-		h *= r_fboResolution.GetFloat();
-	}
-	qglScissor( x, y, w, h );
+	frameBuffers->activeFbo->SetScissorVidSize( x, y, w, h );
```
stgatilov: I understand why scissor and viewport setters are non-static methods of FBO. But I think this sort-of implies that they are the part of FBO while they are not. When different FBO is bound, caller must reset viewport and scissor. This is easy to forget.

I wonder if it is possible to make scissor and viewport part of FBO state?... So that when new FBO is bound, it check if they are the same and send new values to GL driver if they have changed.

renderer/RenderSystem_init.cpp
```diff
@@ -234,14 +235,11 @@ idCVar r_glCoreProfile( "r_glCoreProfile", "2", CVAR_RENDERER | CVAR_ARCHIVE,
 idCVarBool r_newFrob( "r_newFrob", "0", CVAR_RENDERER | CVAR_ARCHIVE, "1 = use the frob shader instead of material stages" );
 
 // FBO
-idCVar r_useFbo( "r_useFBO", "1", CVAR_RENDERER | CVAR_BOOL, "Use framebuffer objects" );
 idCVar r_showFBO( "r_showFBO", "0", CVAR_RENDERER | CVAR_INTEGER, "0-5 individual fbo attachments" );
-idCVar r_fboColorBits( "r_fboColorBits", "32", CVAR_RENDERER | CVAR_INTEGER | CVAR_ARCHIVE, "15, 32, 64" );
+idCVar r_fboColorBits( "r_fboColorBits", "64", CVAR_RENDERER | CVAR_INTEGER | CVAR_ARCHIVE, "32, 64" );
```
stgatilov: Already a default? Huh...

renderer/FrameBuffer.cpp
fholger: Yeah, fair enough :) I'm too used to have a linter and auto-formatter to take care of these things :D

renderer/FrameBuffer.cpp
fholger: You're right, it's not a well-designed interface. Hmm...

Having FBOs manage their individual viewport/scissor state is not the worst idea. The probability that you'd need to (re-)set these when switching FBOs is quite high. I still think a stack-based state management approach to allow localized state changes would catch more errors, but it's worth think about, either way.

renderer/RenderSystem_init.cpp
fholger: Yes. I mean, this is targeted at 2.09, which, by our current release intervals, is quite some time in the future. In that period I think a reasonable amount of people will have used 64 bits with 2.08, and I also plan to do some initial alpha testing with a few people for the FBO and draw call reduction changes to hopefully catch the most glaring issues early. If in that time or in the beta period for 2.09 we do find some major issues with 64 bits, we can revert to 32. But from what I've seen, the performance impact is minimal, and the banding reduction and potential for bloom is significant enough that I think we should aim for 2.09 to default to 64 bits, if at all possible :)

renderer/FrameBuffer.cpp
fholger: Given that there's not really any code left in that file to which this could apply, I just deleted it. (And if it's needed again, it's in the history, anyway.)

renderer/FrameBuffer.cpp
fholger: I just moved the functionality back into external helper functions GL_ViewportXXX and GL_ScissorXXX. In the end, there wasn't really any need to have them be member functions in the first placed.

PR #20, opened 2020-06-13
New render backend with multidraw/bindless support

fholger:
TODO:
* shadowmap support in interaction stage
* (optional) shadow mapping stage?


engine.vcxproj
```diff
@@ -1,1479 +1,1500 @@
...
-    </ProjectReference>
-    <ProjectReference Include="typeinfo.vcxproj">
-      <Project>{6ea6406f-3e65-47d9-8246-d6660a81606f}</Project>
-      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
-    </ProjectReference>
-  </ItemGroup>
-  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-  <ImportGroup Label="ExtensionTargets">
-  </ImportGroup>
+<?xml version="1.0" encoding="utf-8"?>
```
stgatilov: Ooops =)
What happened to this one?

Converting EOLs to Linux-style may be a problem. I wonder if editorconfig settings take effect...

glprogs/stages/depth/depth.frag.glsl
```diff
@@ -0,0 +1,46 @@
...
+out vec4 FragColor;
+
+#ifdef BINDLESS_TEXTURES
+vec4 textureAlpha() {
+    sampler2D tex = sampler2D(params[var_DrawId].texture);
+    return texture(tex, var_TexCoord0.st);
+}
+#else
+uniform sampler2D u_texture;
+
```
stgatilov: (perhaps remove this empty line?)

glprogs/stages/depth/depth.frag.glsl
```diff
@@ -0,0 +1,46 @@
...
+    return texture(u_texture, var_TexCoord0.st);
+}
+#endif
+
+void main() {
+	if (clipPlaneDist < 0.0)
+		discard;
+	vec4 scissor = params[var_DrawId].scissor;
+	// manual scissor test to avoid GL state change; needed to cull invisible parts of surfaces around visportals
+	if (gl_FragCoord.x < scissor.x || gl_FragCoord.y < scissor.y || gl_FragCoord.x >= scissor.z || gl_FragCoord.y >= scissor.w)
```
stgatilov: We have glScissor which uses [x..x+w) x [y..y+h) convention, engine scissors which have [x1..x2] x [y1..y2] convention, and now also this code using [x1..x2) x [y1..y2) convention. While I hate the both-inclusive convention that engine uses, I fear that if you introduce new convention, then we will have some off-by-one-pixel checks in future...

I feel that this confusion is strenghtened by the fact that `scissor` is float-valued.
If other shaders had this test too, then moving it to common header would be OK, but for some reason other shaders don't have it...

glprogs/stages/depth/depth.params.glsl
```diff
@@ -0,0 +1,14 @@
+#pragma tdm_define "MAX_SHADER_PARAMS"
+
+struct ShaderParams {
+    mat4 modelViewMatrix;
+	mat4 textureMatrix;
+	vec4 color;
+	vec4 scissor;
+	vec2 alphaTest;
```
stgatilov: Could you please elaborate why it is 2D?

glprogs/stages/depth/depth.params.glsl
```diff
@@ -0,0 +1,14 @@
+#pragma tdm_define "MAX_SHADER_PARAMS"
+
+struct ShaderParams {
+    mat4 modelViewMatrix;
+	mat4 textureMatrix;
```
stgatilov, line 5: The code has some indentation problems.

Looking through the existing code, I must say that it has much more indentation problems. Perhaps your new backend is the right time to do normalization.

glprogs/stages/depth/depth.vert.glsl
```diff
@@ -0,0 +1,26 @@
...
+};
+
+#pragma tdm_include "stages/depth/depth.params.glsl"
+
+uniform vec4 u_clipPlane;
+uniform mat4 u_inverseView;
+
+out float clipPlaneDist; 
+out vec4 var_TexCoord0;
+flat out int var_DrawId;
```
stgatilov, line 18: Wow, that's an example of ill-chosen naming convention: flat **var**ying =)

Seriously though, I think now we have to stick with it. Number of flat variables won't be high, I hope.

glprogs/stages/interaction/interaction.ambient.fs.glsl
```diff
@@ -0,0 +1,139 @@
...
+in vec4 var_ClipPosition;
+flat in int var_DrawId;
+
+out vec4 FragColor;
+     
+#ifdef BINDLESS_TEXTURES
+vec4 textureNormal(vec2 uv) {
+    sampler2D normalTexture = sampler2D(params[var_DrawId].normalTexture);
+    return texture(normalTexture, uv);
+}
```
stgatilov, line 29: I wonder if we should wrap this into a macro.
Something like `BTSAMPLE(texname, ...)`.

glprogs/stages/interaction/interaction.stencil.fs.glsl
```diff
@@ -0,0 +1,121 @@
...
+
+uniform usampler2D u_stencilTexture;
+uniform sampler2D u_depthTexture;
+
+uniform int u_testStencilSelfShadowFix;
+in vec3 var_WorldLightDir;
+
+out vec4 FragColor;
+
+vec2 renderResolution = vec2(textureSize(u_depthTexture, 0));
```
stgatilov: Just a minor note: we had a bug on AMD when sampling texture during global initialization yielded zero. To be safe, better make this a local variable.

glprogs/stages/interaction/interaction.stencil.fs.glsl
```diff
@@ -0,0 +1,121 @@
...
+	return B / (A + clipZ);
+}
+
+void StencilSoftShadow() {
+	vec2 texSize = renderResolution;
+	vec2 pixSize = vec2(1.0, 1.0) / texSize;
+	vec2 baseTC = gl_FragCoord.xy * pixSize;
+
+	float StLevel = 129.0;
+	if (u_testStencilSelfShadowFix != 0) {
```
stgatilov, line 27: I don't recall if anyone actually tested it. But I find this "hack" too bad to be used.
I'd say it should be removed, although now it's probably etter to wait until these changes get merged.

renderer/GLSLProgram.cpp
```diff
@@ -337,6 +353,19 @@ GLuint GLSLProgram::CompileShader( GLint shaderType, const char *sourceFile, con
 	return shader;
 }
 
+void GLSLProgram::SetDefaultUniformBlockBindings() {
+	GLuint projMatrixBlockIndex = qglGetUniformBlockIndex( program, "block" );
```
stgatilov: "block"? "ShaderParamsBlock" is not much better though...
Is it possible to create a better name?

Maybe something like global and per-object data, I don't know. It depends on your vision of how each of the blocks can and can't grow in future.

renderer/Image_init.cpp
```diff
@@ -238,6 +238,8 @@ idImage::idImage() {
 	memset( &cpuData, 0, sizeof( cpuData ) );
 	residency = IR_GRAPHICS;
 	backgroundLoadState = IS_NONE;
+	isGpuResident = false;
+	textureHandle = 0;
```
stgatilov, line 242: `lastNeededInFrame` should be left uninitialized?

renderer/Image_load.cpp
```diff
@@ -1708,3 +1709,50 @@ void idImage::Print() const {
 	}
 	common->Printf( " %s\n", imgName.c_str() );
 }
+
+void idImage::MakeResident() {
+	if (residency & IR_GRAPHICS == 0) {
```
stgatilov: Bitwise operators should be surrounded by parentheses.

This time precedence is wrong for sure =)

renderer/Image.h
```diff
@@ -256,6 +256,16 @@ class idImage {
 	imageBlock_t		cpuData;				// CPU-side usable image data (usually absent)
 	imageResidency_t	residency;				// determines whether cpuData and/or texnum should be valid
 	imageLoadState_t	backgroundLoadState;	// state of background loading (usually disabled)
+
+	// START bindless texture support
+private:
+	GLuint64			textureHandle;
+	bool				isGpuResident;
```
stgatilov: It seems we have two "resident" members now.
Would it make sense to combine them or rename one of them?

renderer/backend/UniversalGpuBuffer.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+******************************************************************************/
+#include "precompiled.h"
+#include "UniversalGpuBuffer.h"
+#include "../tr_local.h"
+
+extern idCVarBool r_usePersistentMapping;
+
+void UniversalGpuBuffer::Init( GLenum type, GLuint size, GLuint alignment ) {
+	if( mMapBase ) {
```
stgatilov: `mBufferObjects` looks like a better candidate.

renderer/backend/UniversalGpuBuffer.cpp
```diff
@@ -0,0 +1,158 @@
...
+	mAlign = alignment;
+	mType = type;
+
+	qglGenBuffers( 1, &mBufferObject );
+	qglBindBuffer( type, mBufferObject );
+
+	mUsePersistentMapping = r_usePersistentMapping && glConfig.bufferStorageAvailable;
+
+	if (mUsePersistentMapping) {
+		qglBufferStorage( type, mSize, nullptr, GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT );
```
stgatilov: I wonder what do you think about storing static data in same buffer. I guess otherwise you have to rebind different buffers between draw calls. Or have you avoided it somehow?

To be honest, putting everything to one buffer is ugly. But performance-wise, I'm afraid it had some positive effect.

renderer/backend/UniversalGpuBuffer.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+extern idCVarBool r_usePersistentMapping;
+
+void UniversalGpuBuffer::Init( GLenum type, GLuint size, GLuint alignment ) {
+	if( mMapBase ) {
+		Destroy();
+	}
+
+	mSize = ALIGN( size, alignment );
+	mAlign = alignment;
```
stgatilov: assert for power of two?

renderer/backend/UniversalGpuBuffer.h
```diff
@@ -0,0 +1,55 @@
...
+ and/or modify it under the terms of the GNU General Public License as
+ published by the Free Software Foundation, either version 3 of the License,
+ or (at your option) any later version. For details, see LICENSE.TXT.
+
+ Project: The Dark Mod (http://www.thedarkmod.com/)
+
+******************************************************************************/
+#pragma once
+
+class UniversalGpuBuffer {
```
stgatilov: I'm sorry but I would ask you to wrote a comment here about generic idea of this class (maybe comments for some methods too). It takes serious time to read all methods and understand what's going on.

For instance:

 * intended size of buffer?  (pretty large, I guess)
 * pointer arguments?  (all point to CPU-side buffer, which is mirrored in GPU, the buffer is never resized/reallocated)
 * how data is filled? overflow handling? (looks like circular buffer, not sure about overflow)
 * reserve/commit interaction?
 * when is Lock method used?

renderer/backend/UniversalGpuBuffer.cpp
```diff
@@ -0,0 +1,158 @@
...
+	assert( mapOffset >= 0 && mapOffset < mSize );
+
+	// for persistent mapping, nothing to do. Otherwise, we need to upload the committed data
+	if( !mUsePersistentMapping ) {
+		qglBindBuffer( mType, mBufferObject );
+		qglBufferSubData( mType, mapOffset, requestedSize, offset );
+	}
+
+	// mark as used
+	mCurrentOffset = std::max( static_cast< GLuint >( mapOffset + requestedSize ), mCurrentOffset );
```
stgatilov: Is it correct to commit arbitrary range without reserving it first?

renderer/backend/UniversalGpuBuffer.h
```diff
@@ -0,0 +1,55 @@
...
+	void Commit( byte *offset, GLuint size );
+	void BindRange( GLuint index, byte *offset, GLuint size );
+	void Bind();
+	const void * BufferOffset( const void *offset );
+
+private:
+	struct LockedRange {
+		GLuint offset;
+		GLuint count;
+		GLsync fenceSync;
```
stgatilov: Perhaps write a destructor which asserts for zero `fenceSync`?

It is not easy to verify that these objects don't leak, and I fear somebody can break it in future.

renderer/backend/UniversalGpuBuffer.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+byte * UniversalGpuBuffer::Reserve( GLuint size, bool precommit ) {
+	GLuint requestedSize = ALIGN( size, mAlign );
+	if( requestedSize > mSize ) {
+		common->Error( "Requested shader param size exceeds buffer size" );
+	}
+
+	if( mCurrentOffset + requestedSize > mSize ) {
+		if( mCurrentOffset != mLastLocked ) {
+			LockRange( mLastLocked, mCurrentOffset - mLastLocked );
```
stgatilov: I don't get what it does.

As far as I understand, user 1) allocates and writes some data, 2) renders this date, 3) calls `Lock` to mark the data as to-be-free when rendering issued before this moment is over.

Now let's consider two situations:

 * *Wraparound*. Let's suppose I reserve 100 allocations, and the 74-th allocation doesn't have enough space at the end of the buffer. Then the reserved 73 allocations are locked with a fence, although I have not issued draw command for them yet. (BTW, is it possible that I haven't even put my data into them yet?) So these 73 allocations can get overwritten with new data before being rendered if the buffer moves fast enough.

 * *Overflow*. Let's suppose I reserve so many allocations that their total size is larger than size of whole buffer. While the same wraparound issue can happen here, I also have perfect guarantee that I would overwrite my own data. And there seems to be no check for it, no way to detect the situation.

Thinking about it, it seems more correct like this:

 * Don't lock range here, and don't reset `mLastLocked` to zero. Just like in an ordinary circular buffer, allow `beg > end` situation. Handle it explicitly in `Lock`, i.e. issue two syncs if it happens.

 * Detect overflow when `end < beg && end + addedSize >= beg`. No idea what to do in such case, since I don't understand the whole usage pattern yet =) Perhaps return NULL, so that user can issue pending draw commands and retry.

renderer/tr_main.cpp
```diff
@@ -1027,6 +1027,10 @@ static int R_QsortSurfaces( const void *a, const void *b ) {
 	if ( ea->sort > eb->sort ) {
 		return 1;
 	}
+	// sort by material to reduce texture state changes in depth stage
```
stgatilov, line 1031: I thought depth stage does not use any textures... Is this comment still relevant?

renderer/tr_render.cpp
```diff
@@ -327,6 +328,7 @@ void RB_EnterWeaponDepthHack() {
 	matrix[14] *= 0.25f;
 
 	if ( r_uniformTransforms.GetBool() ) {
+		// FIXME: this part is broken since storing projection matrix in uniform block
```
stgatilov, line 331: I suggest asking Duzenko [here](https://forums.thedarkmod.com/index.php?/topic/20096-modelweapon-depth-hack/).

As far as I see it, he already disabled it before beta started, and nobody noticed.

renderer/backend/DrawBatchExecutor.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void DrawBatchExecutor::Destroy() {
+	qglDeleteBuffers(1, &drawIdBuffer);
+	drawIdBuffer = 0;
+	drawCommandBuffer.Destroy();
+}
+
+void DrawBatchExecutor::BeginBatch( int maxDrawCalls ) {
+	maxDrawCommands = maxDrawCalls;
```
stgatilov: A check that it does not exceed MAX_DRAW_COMMANDS?

renderer/backend/DrawBatchExecutor.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void DrawBatchExecutor::InitDrawIdBuffer() {
+	qglGenBuffers(1, &drawIdBuffer);
+	qglBindBuffer(GL_ARRAY_BUFFER, drawIdBuffer);
+	std::vector<uint32_t> drawIds (MAX_DRAW_COMMANDS);
+	for (uint32_t i = 0; i < MAX_DRAW_COMMANDS; ++i) {
+		drawIds[i] = i;
+	}
+	qglBufferData(GL_ARRAY_BUFFER, drawIds.size(), drawIds.data(), GL_STATIC_DRAW);
```
stgatilov: I think size should be multiplied by 4.

Or use `MemoryUsed`? =)))

renderer/backend/DrawBatchExecutor.cpp
```diff
@@ -0,0 +1,167 @@
...
+	qglBindBuffer(GL_ARRAY_BUFFER, 0);
+
+	qglVertexAttribIFormat( Attributes::Default::DrawId, 1, GL_UNSIGNED_INT, 0 );
+	qglBindVertexBuffer( Attributes::Default::DrawId, drawIdBuffer, 0, sizeof(uint32_t) );
+	qglVertexAttribBinding( Attributes::Default::DrawId, Attributes::Default::DrawId );
+	qglVertexBindingDivisor( Attributes::Default::DrawId, 1 );
+	qglEnableVertexAttribArray( Attributes::Default::DrawId );
```
stgatilov, line 126: What an abomination!
Just wanted to clarify: is it really necessary here?

renderer/backend/DrawBatchExecutor.h
```diff
@@ -0,0 +1,55 @@
...
+class DrawBatchExecutor {
+public:
+	void Init();
+	void Destroy();
+
+	void BeginBatch(int maxDrawCalls);
+	void AddDrawVertSurf(const drawSurf_t *surf);
+	void AddShadowSurf( const drawSurf_t * surf );
+	void DrawBatch();
+	void Lock();
```
stgatilov: It is not clear what this method is for.
A comment would be welcome.

Logically, `DrawBatch` can `Lock` internally at the end of the function...

stgatilov (commented):
I'm sorry, but I did not finish it in one day =)

I have not looked at:
 * all stages (C++ code)
 * RenderBackend
 * VertexCache

Nevertheless, I decided to push this portion for now.

engine.vcxproj
fholger: Yeah, no idea what happened, 3 files were randomly affected, and I don't know why. But I'm already in the process of fixing it...

fholger: I repaired the files, sorry.

glprogs/stages/depth/depth.frag.glsl
fholger: You're right, it shouldn't be float, my mistake. Apart from that the values passed and tested against should follow the same principles as our previous use of glScissor, but I'll double-check.

Other shaders don't have it due to depth test - if the depth stage has rejected the unwanted pixels, they will fail the depth test in later stages.

Just to be clear, I'm not a particular fan of the manual scissor test. But it is the only way to batch draw calls and avoid a GL state change :(

glprogs/stages/depth/depth.params.glsl
fholger: UBOs have very strict alignment rules, so I can't put a single float here, because the entire struct would then not adhere to those rules. (Or actually, here in the shader I could, but it would add implicit padding which you'd need to account for the mirrored struct in the C++ code, which I think would be even more confusing.)

But on second thought, it'd probably be better to use two floats and name the second one `padding` or something similar.

renderer/backend/DrawBatchExecutor.h
fholger: Each `Lock` issues a glFenceSync, which are expensive. Yes, you could do it after every DrawBatch, but performance would suffer, so I do it explicitly only at the end of a frame (or implicitly when the buffer wraps around). I can add some comments.

glprogs/stages/depth/depth.vert.glsl
fholger: Honestly didn't think about it :)
But hm, this only affects my new stage shaders, so there's ample opportunity here to name them differently, or just drop the `var_` ...

renderer/tr_main.cpp
fholger: It does unfortunately, for alpha tests in perforated materials...

glprogs/stages/depth/depth.frag.glsl
fholger: I changed scissor to a uvec4 and changed its meaning to x, y, width, height, which should now more closely resemble other uses in the engine.

glprogs/stages/interaction/interaction.stencil.fs.glsl
fholger: Agreed. I think there are a number of cvars that should be removed after the changes are merged - some obvious ones like r_uniformTransforms and r_useGlsl, and some where we should decide on a behaviour, like this one and the specular fix.

renderer/GLSLProgram.cpp
fholger: `block` is unfortunately a term that duzenko chose in the existing shaders, and I didn't want to risk changing them and breaking the old renderer in the process. I have gone with `ViewParamsBlock` in the new shaders. And fair enough, I could change `ShaderParamsBlock` to `PerObjectParamsBlock`, perhaps?

renderer/Image_init.cpp
fholger: Nope, just an oversight.

renderer/Image.h
fholger: Unfortunately, that is the nomenclature with bindless textures, you have to make them "resident" before using them in a shader. I think it's potentially error-prone to combine it with the existing enum, but I'll rename the variable `isBindlessHandleResident`, which should hopefully make it clearer that it's a different kind of residency.

glprogs/stages/depth/depth.frag.glsl
stgatilov: Actually, depth test does not guarantee correctness. Suppose that a triangle goes out of the scissor region and intersects another surface (which was drawn with different scissor). When this triangle is rendered with depth=equal condition, some pixels can accidentally have exactly the same depth, and they will overwrite color.

However, I think correct scenes should not have anything out of the scissor (except for shadow volumes), so if I won't worry about it until someone finds a real case.

glprogs/stages/depth/depth.vert.glsl
stgatilov: In my opinion it's better to stick to var, or rename everything at once.

renderer/backend/DrawBatchExecutor.cpp
fholger: Yep, good catch. That would have made for a nasty bug if we ever reach more than 1024 draw calls in a single batch :D

renderer/backend/DrawBatchExecutor.cpp
fholger: What exactly do you mean? The setup is necessary unfortunately, though I suppose I could move it to the RenderBackend. Don't really have any strong opinions on that :)

renderer/GLSLProgram.cpp
stgatilov: `PerObjectParamsBlock` sounds better. Maybe `draw` would be better that `object`, but not sure.

glprogs/stages/depth/depth.frag.glsl
fholger: Right, but we'd really have to get unlucky to encounter something like that by accident, and I really hope no mapper would set up such a scenario on purpose :D So I'm with you, wait until it's an actual problem and then deal with it.

renderer/Image.h
stgatilov: Actually, it is not too different in theory if you think about it:

 * CPU-resident: cpu-side buffer is valid
 * GPU-resident: texture "name" is valid  (is it the right GL term? =) )
 * the-new-resident: texture bindless handle is valid

But in practice it may be not so clear. Anyway, I don't argue with leaving it as it is now, it is pretty clear what it means.

renderer/backend/UniversalGpuBuffer.cpp
fholger: At the moment, this class is used to store per drawcall params and multidraw commands for drawcall batching, so no static data.

Static data only becomes an issue if we want to migrate the VertexCache to also use this `UniversalGpuBuffer`. I think it would be a good idea, but this PR is already huge, so let's postpone that :)

And yeah, in general you'd want to use fewer buffers, but it's primarily about batching draw calls without state change - if static objects were still in a different buffer (before duzenko made those changes), then you'd have to accomodate for that in setting up the draw call batches in the new stages. It would be doable and wouldn't really be much of a performance problem if you ensure that you only switch the static/dynamic buffers once, but it does make the code more tedious.

glprogs/stages/depth/depth.params.glsl
stgatilov: Padding sounds better. Some people also use names like `_` or `_1`, `_2`, etc. It would immediately show the reason for all of this. The problem of different layout will hunt us all the time now, so reader has know about it.

2D vector with unused sounds more confusing. Reminds ARB shaders with their "everything is vec4" rule =)

fholger: Yep. I've already put in place a STATIC_ASSERT that will tell you if your params struct is not a multiple of 16, but this doesn't catch all the mistakes you can do. Still better than nothing, I think.

glprogs/stages/depth/depth.params.glsl
fholger: I reformatted the new shaders.

glprogs/stages/depth/depth.vert.glsl
fholger: Alright. :)

glprogs/stages/interaction/interaction.stencil.fs.glsl
fholger: Good to know, I inlined it in the function.

renderer/backend/UniversalGpuBuffer.h
fholger: `Destroy` already clears any remaining fence syncs, not sure what the assert would accomplish?

renderer/backend/UniversalGpuBuffer.cpp
fholger: I'll add the requested comments to the class, which will hopefully make it clearer. But I do think the behaviour as it is is correct, as long as you keep to the contract of the interface - which unfortunately is a little awkward, but I haven't been able to find anything better.

The contract is basically:

1. You request a certain amount of storage from the buffer. It cannot immediately be marked as used for two reasons, one being the fallback from persistent buffers and the other being the backend design where it's virtually impossible to know how many draw calls you'll have in a batch until you actually start assembling them.
2. Once you've filled the requested memory with your data, and are ready to use it, you commit the data, which will mark it as used and, in case of fallback, actually upload it (not needed with persistent mapping).
3. Now you should use the data in draw calls before requesting any new storage.
4. Finally, this is where you could lock the used range. However, since fence syncs are expensive, we can't lock after every small draw batch, so by default a Lock is only issued once at the end of a frame. The only exception is if the buffer needs to wrap around.

If you stick with this contract, both of your outlined situations cannot happen - the locking will specifically ensure that overflow is not a problem, unless you make a single request that is larger than the entire buffer. If you don't stick with it, you can get into even more problems, I think. Unfortunately, I don't see a good way to enforce this contract or provide a better interface. I'm open to ideas. I do agree it needs documentation.

renderer/backend/UniversalGpuBuffer.cpp
fholger: Well, the compiler will accept it, but it's not semantically correct (see explanation below).

glprogs/stages/interaction/interaction.ambient.fs.glsl
fholger: My feeling is that for the current number of occurrences, it's easier to read without the macro. If we do write additional shaders with additional instances of this, then I'd revisit that suggestion. I don't see obvious candidates right now, though.

renderer/backend/UniversalGpuBuffer.h
stgatilov: Not all structs are removed in `Destroy`, usually they are removed during normal operation. Right now one has to manually check that every such place deletes the fence. I imagine with a bit of changes those fences can start leaking, and that would be hard to notice. The assert would prevent that.

renderer/backend/UniversalGpuBuffer.cpp
stgatilov: The fences are used to ensure that when you wraparound, you wait for GL commands using the old data to complete before writing the new data. For it to happen, the fence must be issued after the **user has finished all rendering with the data**. Issuing fence without explicit `Lock` creates race condition, because the user does not know about it, and he issues draw commands after the fence.

renderer/backend/UniversalGpuBuffer.cpp
stgatilov: Perhaps add a check for it then?
This situation can happen especially when calling commit on stale pointer.

I think it is as simple as checking that offset falls between `mLastLocked` and `mCurrentOffset`.

renderer/backend/DrawBatchExecutor.h
stgatilov: Add comment for the Lock method at least, please.

renderer/backend/UniversalGpuBuffer.cpp
stgatilov: Speaking of the *overflow* situation, it seems that it is not supported. I.e. if it happens, geometry will disappear randomly or just render incorrectly. I'm afraid in order to support overflow, the user must know when his not-yet-locked allocations has filled the whole buffer, so that he could issue draw commands before pushing more. There is no interface for it, and it is not easy to do, since requires assistance from the user.

Note that I'm not speaking about the case when single allocation is larger than buffer --- that's clearly forbidden and even checked for.

fholger: The contract is that you must have committed *and rendered* any previous data before you request new data from the buffer, that's the contract. Therefore, the issuing of the fence is correct because the as-yet unlocked data *has been used in a draw call* per the contract. And that's also why the overflow works - you land back at the beginning of a buffer and then explicitly wait for any fences that overlap with the requested range - including, theoretically, the range you only just locked (which will guarantee a wait, obviously).

Obviously, you can break that contract, like request some data, fill it, then commit it and immediately request another block and only issue draw calls after both. And this is a scenario that I wish I could support, because there are possibly instances where it could be useful to have prepared multiple blocks for a single draw call. But it is simply forbidden due to the very reason you outlined.

Again, if you have ideas for a better interface, I'm open to it. I've thought about several alternatives, and in the end, they've all been bad for various reasons.

renderer/backend/UniversalGpuBuffer.h
fholger: Perhaps, but you'll never arrive in the destructor before first having called `Destroy` - at least not without yet another bug. So the assert in the destructor will not help. And you can't place the assert in `Destroy` because having some fences still pending is not an error.

stgatilov: `WaitForLockedRange` destroys range directly. And it produces a feeling that it is perfectly OK to just write `mRangeLocks.erase(mRangeLocks.begin())` somewhere.

renderer/backend/UniversalGpuBuffer.cpp
fholger: One rough idea I had is I could perhaps fuse the DrawBatchExecutor and the ShaderParamsBuffer together. Right now, they are independent abstractions on top of the UniversalGpuBuffer, but you'd basically always use them in tandem. And if that's reflected in their combined interface, it would be much harder to misuse them. You could still misuse the UniversalGpuBuffer, but that's a lower primitive that you'd not normally use directly.

renderer/backend/UniversalGpuBuffer.h
fholger: I'm still not quite sure what it is you want me to do? Yes, the class is delicate because it manages GPU resources - if you alter it without knowing what you're doing you're going to cause issues. And I'm happy with whatever check catches potential errors, but the assert in the destructor will simply not do anything.

Only alternative I see is to abandon the concept of a circular buffer entirely and instead make it a "triple-range" buffer that uses one explicit range per frame. It would simplify the locking concept and abandon any issues with wraparound. On the downside it would mean that the buffer needs to be large enough to hold all data for the frame data of the worst-case scene with no wiggle room to deal with an occasional spike.

renderer/backend/UniversalGpuBuffer.cpp
stgatilov: I think it's better to have usage patterns of this class crystal clear. It is usually easier to understand something smaller, than the whole mix of three classes. And you said that you think about migrating `VertexCache` to this buffer, so it is important to make it clear.

Ok, I see it now: user can only Reserve+Commit+Draw+Reserve+Commit+Draw+... in this strict sequence, and Lock is optional and used only for better performance. Basically, only one allocation can be alive at any moment. Sounds like a consumer of immediate-style data, no reuse between allocations is allowed.
How can VertexCache be implemented then? Would you sort all surfaces by vertex pointer to ensure that all draws of the same mesh go at the same time, or load the mesh many times?

As for suggestions, I made one in the original post. You can allow the situation when `mLastLocked > mCurrentOffset`, making it fully circular. If it happens, then you have to issue two fences in `Lock` instead of one, but that's not a big problem. If you do so, then the user can reserve several allocations in a row, then commit them in arbitrary order, then draw various stuff using them in arbitrary order, and finally `Lock`. Of course, 1) locking would become mandatory, and 2) total size of unlocked allocations must not exceed buffer size.

Regarding the last condition, I think it can be lifted away if you return NULL on allocation with overflow. If the user issues all draws before allocating more stuff (as you expect it now), then it is enough for him to check for NULL, and call `Lock`  + retry `Reserve` if it happened. But for something like VertexCache, caller can resize the whole buffer for instance.

I'm not sure if it is better. I guess this question would not appear if the class had a comment from the very beginning.

fholger: After this discussion, I'm beginning to favour the approach of abandoning a circular buffer and using the 'triple-range' concept loosely outlined in the other thread. Combined with an overhaul of the DrawBatchExecutor/ShaderParamsBuffer, it could potentially result in a cleaner interface with fewer chances of misuse. And you're right, the circular pattern right now would have some issues to be used in the VertexCache...

But it's not a trivial change, so I'll have to think about it and experiment a little.

fholger: I have committed a reworked `GpuBuffer` and `DrawBatchExecutor`, the latter also taking over and replacing the responsibilities of `ShaderParamsBuffer`. I also added some comments about its workings.

As proposed, the new `GpuBuffer` is no longer a circular buffer, but rather divided into three frame regions (and an optional static data region), where each frame writes to particular region, and then at the end of the frame that region is locked as a whole and you switch to the next region. It should make the whole thing conceptually simpler, as there's no wraparound to worry about.

The `DrawBatchExecutor` will now give you a simple struct for a DrawBatch with two arrays that you have to fill with the shader params and the drawSurfs you want to draw and will handle the `GpuBuffer` for the UBO automatically in the background for you. It still doesn't have the greatest interface, but I think hopefully it is a bit clearer and much harder to misuse. Let me know what you think.

renderer/backend/UniversalGpuBuffer.h
fholger: As proposed, I've abandoned the circular buffer, so this circular locking concept is no longer needed.

renderer/backend/UniversalGpuBuffer.cpp
fholger: New `GpuBuffer` has a simpler `Commit` interface, so this should hopefully not be an issue anymore.

renderer/backend/GpuBuffer.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+	qglGenBuffers( 1, &bufferObject );
+	qglBindBuffer( type, bufferObject );
+
+	usesPersistentMapping = r_usePersistentMapping && GLAD_GL_ARB_buffer_storage;
+
+	totalSize = staticDataSize + NUM_FRAMES * frameSize;
+	if ( usesPersistentMapping ) {
+		qglBufferStorage( type, totalSize, nullptr, GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT );
+		bufferContents = ( byte* )qglMapBufferRange( type, 0, totalSize, GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT );
```
stgatilov, line 42: A stupid question came to my mind, related to `GL_MAP_COHERENT_BIT`.

If you map the whole buffer forever, then only write to it using CPU instructions, how does driver understand what it should send to GPU and when? I mean: in a typical situation GPU memory and CPU mapped memory are different locations, and copying should occur.

 1. Is it certain that driver won't copy the whole buffer every frame instead of one third of it?
 2. Is it certain that driver will copy contents when you have finished writing them, and not when you wait for fence before drawing them?

renderer/backend/GpuBuffer.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+byte * GpuBuffer::CurrentWriteLocation() const {
+	return bufferContents + CurrentOffset();
+}
+
+GLuint GpuBuffer::BytesRemaining() const {
+	return frameSize - bytesCommittedInCurrentFrame;
+}
+
+void GpuBuffer::Commit( GLuint numBytes ) {
```
stgatilov, line 89: Suppose user wants to put many draw surfaces into vertex cache.
Then he has either to manage alignment himself, or to call `CurrentWriteLocation` + `Commit` for every surface.

In the second case, would it have tolerable performance on old hardware with many calls of `glBufferSubData` ?
Will it work faster if `glBufferSubData` was called once on a big memory chunk?

renderer/backend/GpuBuffer.cpp
```diff
@@ -0,0 +1,144 @@
...
+	GLintptr mapOffset = offset - bufferContents;
+	assert(mapOffset >= 0 && mapOffset < totalSize);
+	qglBindBufferRange( type, index, bufferObject, mapOffset, size );
+}
+
+void GpuBuffer::Bind() {
+	qglBindBuffer(type, bufferObject);
+}
+
+const void * GpuBuffer::BufferOffset( const void *offset ) {
```
stgatilov: I think parameter should be renamed from `offset` to `pointer`

renderer/backend/GpuBuffer.cpp
```diff
@@ -0,0 +1,144 @@
...
+		qglBufferSubData( type, currentOffset, alignedSize, bufferContents + currentOffset );
+	}
+
+	bytesCommittedInCurrentFrame += alignedSize;
+}
+
+void GpuBuffer::BindRange( GLuint index, byte *offset, GLuint size ) {
+	GLintptr mapOffset = offset - bufferContents;
+	assert(mapOffset >= 0 && mapOffset < totalSize);
+	qglBindBufferRange( type, index, bufferObject, mapOffset, size );
```
stgatilov, line 106: I am a bit confused with `Bind` vs `BindRange`. It looks like `BindRange` can only be used for some specific types of buffers like UBO, but not for vertex buffers. And `Bind` cannot specify `offset` and `size`. By the way, why is `size` needed here?

Maybe rename the methods? Current names suggest that main difference is that this one allows to bind a part of the buffer, but the main difference actually is that this one binds some specific types of buffers. Maybe call it `BindToBindingIndex` ? Although vertex buffers can also be bound to binding indices without yet another set of functions =(

renderer/backend/DrawBatchExecutor.h
```diff
@@ -0,0 +1,103 @@
...
+
+	template< typename ShaderParams >
+	uint MaxShaderParamsArraySize() {
+		return maxUniformBlockSize / sizeof( ShaderParams );
+	}
+
+private:
+	static const uint MAX_SHADER_PARAMS_SIZE = 512;
+	
+    GpuBuffer shaderParamsBuffer;
```
stgatilov: indent

renderer/backend/DrawBatchExecutor.h
```diff
@@ -0,0 +1,103 @@
...
+ * submitted in a single MultiDrawIndirect call or in separate
+ * DrawElementsBaseVertex calls. To pass draw call parameters to the shader,
+ * the class also manages and provides a UBO that can be filled with the
+ * relevant data.
+ *
+ * To start a batch, call `BeginBatch` with the ShaderParams struct that
+ * represents your UBO params structure. Note that this struct must adhere
+ * to the std140 layout rules specified in the OpenGL specification. You
+ * will receive a DrawBatch struct that contains two arrays - one for your
+ * ShaderParams and one for the drawSurfs to render. Fill these arrays up
```
stgatilov: I guess it's better to clarify how exactly user should "fill" the `DrawBatch::surfs`. Something like "assign pointers to your drawSurf-s into batch.surfs[*]". I spent some time to understand why the two other understandings of "fill" contradict with the signature =)

renderer/backend/DrawBatchExecutor.h
```diff
@@ -0,0 +1,103 @@
...
+	template< typename ShaderParams >
+	DrawBatch<ShaderParams> BeginBatch();
+
+	void ExecuteDrawVertBatch( int numDrawSurfs, GLuint uboIndex = DEFAULT_UBO_INDEX );
+	void ExecuteShadowVertBatch( int numDrawSurfs, GLuint uboIndex = DEFAULT_UBO_INDEX );
+
+	void EndFrame();
+
+	template< typename ShaderParams >
+	uint MaxShaderParamsArraySize() {
```
stgatilov: const?

renderer/backend/DrawBatchExecutor.cpp
```diff
@@ -0,0 +1,237 @@
...
+void DrawBatchExecutor::ExecuteShadowVertBatch( int numDrawSurfs, GLuint uboIndex ) {
+	assert( numDrawSurfs <= maxBatchSize );	
+	maxBatchSize = 0;
+
+	byte *shaderParamsContents = shaderParamsBuffer.CurrentWriteLocation();
+	uint shaderParamsCommitSize = numDrawSurfs * shaderParamsSize;
+	shaderParamsBuffer.Commit( shaderParamsCommitSize );
+	shaderParamsBuffer.BindRange( uboIndex, shaderParamsContents, shaderParamsCommitSize );
+
+	vertexCache.BindVertex( ATTRIB_SHADOW );
```
stgatilov: Why do you prefer having all methods duplicated for shadow?

It seems that `ATTRIB_SHADOW` and `sizeof(shadowCache_t)` are the only differences.

stgatilov (commented):
Reviewed new GpuBuffer and DrawBatchExecutor.

glprogs/stages/interaction/interaction.ambient.vs.glsl
```diff
@@ -0,0 +1,57 @@
+#version 330 core
+
+#pragma tdm_include "stages/interaction/interaction.params.glsl"
+
+in vec4 attr_Position;
+in vec4	attr_TexCoord;
```
stgatilov: Ehm... indent? =(

renderer/backend/DepthStage.cpp
```diff
@@ -0,0 +1,336 @@
...
+};
+
+DepthStage::DepthStage( DrawBatchExecutor* drawBatchExecutor )
+	: drawBatchExecutor(drawBatchExecutor)
+{
+}
+
+void DepthStage::Init() {
+	uint maxShaderParamsArraySize = drawBatchExecutor->MaxShaderParamsArraySize<ShaderParams>();
+	depthShader = programManager->LoadFromGenerator( "depth_pass", [=](GLSLProgram *program) { LoadShader(program, maxShaderParamsArraySize, false); } );
```
stgatilov: Is it intended that name of shader object is different from the name of shader file?

I understand the reason for second object and `_bindless` suffix, but wouldn't it be better to either have file named `depth_pass.XXX.glsl` or shader named `depth` ?

renderer/backend/DepthStage.cpp
```diff
@@ -0,0 +1,336 @@
...
+        if ( regs[ pStage->conditionRegister ] != 0 ) {
+            break;
+        }
+    }
+    return stage != shader->GetNumStages();
+}
+
+void DepthStage::DrawSurf( const drawSurf_t *surf ) {
+	if ( surf->space->weaponDepthHack ) {
+		// this is a state change, need to finish any previous calls
```
stgatilov, line 199: It is a bit strange that out of several possible types of state changes, one type (subview) is handled in `DrawDepth`, and all the others are handled in `DrawSurf`.

Wouldn't is make more sense to handle all conditions for state change (and draw calls flushing) in one place?

stgatilov: Ok, I see now that the state changes handled here are always done temporarily for one surface, but state changes for subviews are included into the generic approach of coalescing surfaces together.

renderer/backend/DepthStage.cpp
```diff
@@ -0,0 +1,336 @@
...
+		}
+	}
+
+	if ( drawSolid ) {  // draw the entire surface solid
+		IssueDrawCommand( surf, nullptr );
+	}
+}
+
+void DepthStage::IssueDrawCommand( const drawSurf_t *surf, const shaderStage_t *stage ) {
+	if( stage && !renderBackend->ShouldUseBindlessTextures() && stage->texture.image->texnum != backEnd.glState.tmu[0].current2DMap ) {
```
stgatilov: I wonder if we could have `IsBound` method instead of going to the TMU struct directly?

renderer/backend/DepthStage.cpp
stgatilov: After looking further, I see that there are more checks for state changes in other methods too. I guess it is inevitable =)

renderer/backend/StencilShadowStage.cpp
```diff
@@ -0,0 +1,129 @@
...
+};
+
+StencilShadowStage::StencilShadowStage( DrawBatchExecutor *drawBatchExecutor ) {
+	this->drawBatchExecutor = drawBatchExecutor;
+}
+
+void StencilShadowStage::Init() {
+	idDict defines;
+	defines.Set( "MAX_SHADER_PARAMS", idStr::Fmt("%d", drawBatchExecutor->MaxShaderParamsArraySize<ShaderParams>()) );
+	stencilShadowShader = programManager->LoadFromFiles( "shadows_stencil",
```
stgatilov: Well, the same question: is it normal that shader object names differ from shader file names?

renderer/backend/StencilShadowStage.cpp
```diff
@@ -0,0 +1,129 @@
...
+	stencilShadowShader = programManager->LoadFromFiles( "shadows_stencil", 
+		"stages/stencil/stencilshadow.vert.glsl",
+		"stages/stencil/stencilshadow.frag.glsl",
+		defines );
+}
+
+void StencilShadowStage::Shutdown() {}
+
+void StencilShadowStage::DrawStencilShadows( viewLight_t *vLight, const drawSurf_t *shadowSurfs ) {
+	if ( !shadowSurfs || !r_shadows.GetInteger() ) {
```
stgatilov, line 45: Maybe `r_shadows.GetInteger() != 1` ?

renderer/backend/StencilShadowStage.cpp
```diff
@@ -0,0 +1,129 @@
...
+	qglStencilFunc( GL_ALWAYS, 1, 255 );
+
+	if ( glConfig.depthBoundsTestAvailable && r_useDepthBoundsTest.GetBool() ) {
+		qglEnable( GL_DEPTH_BOUNDS_TEST_EXT );
+	}
+	GL_Cull( CT_TWO_SIDED );
+	stencilShadowShader->Activate();
+
+	std::vector<const drawSurf_t*> depthFailSurfs;
+	std::vector<const drawSurf_t*> depthPassSurfs;
```
stgatilov: I think nothing stops from using `idList` with `AddGrow`.

I wonder if it makes sense to crawl the list with a prepass and then allocate specific size, e.g. using frame temporary memory.

On the other hand, it is done once per every light every frame, it might be that those allocations don't affect performance much.

**EDIT:** I initially thought that stages are throw-away classes, i.e. you create a temporary one when you draw a frame, then drop it when you finish drawing. But it seems that they are actually persistent between frames. Then simplest way to avoid allocations is to save arrays as members.

renderer/backend/GpuBuffer.cpp
fholger: I don't understand the technical details behind it, but yes, GL_MAP_COHERENT_BIT guarantees both.

For point 1, I've seen a bit of the workings due to Nvidia nSight which shows when it's copying. Copies happen right before the first draw call that's accessing state from that buffer. E.g. when you have a certain range of the buffer bound and you issue a draw call, it will trigger an upload. Due to the bound range, it also knows which regions it needs to copy. 
However, I don't know how it knows that the region is even dirty to begin with, but it does - subsequent draw calls don't retrigger the upload :)

renderer/backend/StencilShadowStage.cpp
```diff
@@ -0,0 +1,129 @@
...
+		qglStencilFunc( GL_GEQUAL, 128, 255 ); // enable stencil test - the shadow volume path
+	
+}
+
+void StencilShadowStage::DrawSurfs( const drawSurf_t **surfs, size_t count ) {
+	if (count == 0) {
+		return;
+	}
+
+	vertexCache.VertexPosition( surfs[0]->shadowCache, ATTRIB_SHADOW );
```
stgatilov: Maybe a small comment, since it is not clear what it does...

renderer/backend/GpuBuffer.cpp
fholger: Hm, actually, since the `VertexCache` is filled in the frontend which must not call GL commands at all, we'd need a little extra abstraction there, anyway. So we'd end up with a single `glBufferSubData` per frame out of necessity, and this is what's happening right now, too.

I think that handling could be done entirely in the `VertexCache`; at worst it might need an extra helper function in the `GpuBuffer`.

renderer/backend/InteractionStage.cpp
```diff
@@ -0,0 +1,560 @@
...
+		DEFINE_UNIFORM( int, testStencilSelfShadowFix )
+		DEFINE_UNIFORM( vec4, shadowRect )
+		DEFINE_UNIFORM( int, shadowMapCullFront )
+		DEFINE_UNIFORM( sampler, stencilTexture )
+		DEFINE_UNIFORM( sampler, depthTexture )
+		DEFINE_UNIFORM( sampler, shadowMap )
+	};
+
+	enum TextureUnits {
+		TU_NORMAL = 0,
```
stgatilov, line 85: Wow... so many textures!

I hope we don't use them all at once =)

renderer/backend/DrawBatchExecutor.cpp
fholger: I don't prefer it; I guess the methods ended up more similarly than I originally anticipated. Will try to unify.

renderer/backend/GpuBuffer.cpp
fholger: Hm, the naming is unfortunately already given by GL. You can `Bind` any type of buffer and may have to in certain situations, whereas `glBindBufferRange` only applies to certain types of buffers, which in our case are essentially UBOs only.

The GL docs say "glBindBufferRange — bind a range within a buffer object to an indexed buffer target", so perhaps a better name might be `BindRangeToIndexTarget`?

renderer/backend/InteractionStage.cpp
```diff
@@ -0,0 +1,560 @@
...
+			[this](GLSLProgram *shader) { LoadInteractionShader( shader, "interaction.ambient", true ); } );
+		bindlessStencilInteractionShader = programManager->LoadFromGenerator( "interaction_stencil_bindless", 
+			[this](GLSLProgram *shader) { LoadInteractionShader( shader, "interaction.stencil", true ); } );
+		bindlessShadowMapInteractionShader = programManager->LoadFromGenerator( "interaction_shadowmap_bindless", 
+			[this](GLSLProgram *shader) { LoadInteractionShader( shader, "interaction.shadowmap", true ); } );
+	}
+
+	qglGenBuffers( 1, &poissonSamplesUbo );
+	qglBindBuffer( GL_UNIFORM_BUFFER, poissonSamplesUbo );
+	qglBufferData( GL_UNIFORM_BUFFER, 150 * sizeof(idVec2), nullptr, GL_STATIC_DRAW );
```
stgatilov: Perhaps use `glGetActiveUniformBlockiv` with `GL_UNIFORM_BLOCK_DATA_SIZE` to get the size from shader code?

**EDIT:** I wonder if you really need to initialize buffer here. And if you need, I'm not sure that you need to know the exact number 150 here. Later you set pretty arbitrary size to this buffer.

renderer/backend/InteractionStage.cpp
```diff
@@ -0,0 +1,560 @@
...
+	BeginDrawBatch();
+}
+
+void InteractionStage::PreparePoissonSamples() {
+	int sampleK = r_softShadowsQuality.GetInteger();
+	if ( sampleK > 0 && poissonSamples.Num() != sampleK ) {
+		GeneratePoissonDiskSampling( poissonSamples, sampleK );
+		size_t size = poissonSamples.Num() * sizeof(idVec2);
+		qglBindBuffer( GL_UNIFORM_BUFFER, poissonSamplesUbo );
+		qglBufferData( GL_UNIFORM_BUFFER, size, poissonSamples.Ptr(), GL_STATIC_DRAW );
```
stgatilov, line 557: May I ask you to add a check for overflow?

renderer/backend/InteractionStage.cpp
```diff
@@ -0,0 +1,560 @@
...
+		GL_CheckErrors();
+	}
+
+	// bind the vertex and fragment program
+	ChooseInteractionProgram( vLight, interactionSurfs == backEnd.vLight->translucentInteractions );
+	InteractionUniforms *uniforms = interactionShader->GetUniformGroup<InteractionUniforms>();
+	uniforms->cubic.Set( vLight->lightShader->IsCubicLight() ? 1 : 0 );
+	uniforms->globalLightOrigin.Set( vLight->globalLightOrigin );
+
+	std::vector<const drawSurf_t*> drawSurfs;
```
stgatilov: Should it be a member to avoid allocations?

renderer/backend/DrawBatchExecutor.cpp
```diff
@@ -0,0 +1,237 @@
...
+	if ( r_showPrimitives.GetBool() && !backEnd.viewDef->IsLightGem() && backEnd.viewDef->viewEntitys ) {
+		for ( int i = 0; i < numDrawSurfs; ++i ) {
+			const drawSurf_t *surf = drawSurfs[i];
+			backEnd.pc.c_drawIndexes += surf->numIndexes;
+			if ( surf->frontendGeo ) {
+				backEnd.pc.c_drawVertexes += surf->frontendGeo->numVerts;
+			}
+			backEnd.pc.c_vboIndexes += surf->numIndexes;
+		}
+		backEnd.pc.c_drawElements += ShouldUseMultiDraw() ? 1 : numDrawSurfs;
```
stgatilov: Perhaps we should add a counter for the number of batches sent to driver?

There are so many conditions for state changes which separate stream of surfaces into batches, that it is non-trivial to understand how a change would affect the number of batches.

renderer/backend/RenderBackend.cpp
```diff
@@ -0,0 +1,256 @@
...
+		}
+		DrawShadowsAndInteractions( viewDef );
+	}
+		
+	// now draw any non-light dependent shading passes
+	int RB_STD_DrawShaderPasses( drawSurf_t **drawSurfs, int numDrawSurfs );
+	processed = RB_STD_DrawShaderPasses( drawSurfs, numDrawSurfs );
+
+	// fog and blend lights
+	void RB_STD_FogAllLights( bool translucent );
```
stgatilov: I suggest marking it as extern.
An eye is not trained at all in seeing function declarations in function bodies =)

renderer/backend/RenderBackend.cpp
```diff
@@ -0,0 +1,256 @@
...
+		globalImages->MakeUnusedImagesNonResident();
+	}
+}
+
+bool RenderBackend::ShouldUseBindlessTextures() const {
+	return GLAD_GL_ARB_bindless_texture && r_useBindlessTextures.GetBool();
+}
+
+void RenderBackend::DrawInteractionsWithShadowMapping(viewLight_t *vLight) {
+	void RB_GLSL_DrawInteractions_ShadowMap( const drawSurf_t *surf, bool clear = false );
```
stgatilov: Could you please remove default argument here?

Using local function declarations with default arguments can seriously mess things up if default arguments don't match.

renderer/backend/RenderBackend.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+	GLSLProgram::Deactivate();
+}
+
+void RenderBackend::DrawShadowsAndInteractions( const viewDef_t *viewDef ) {
+	GL_PROFILE( "LightInteractions" );
+
+	if ( r_shadows.GetInteger() == 2 ) {
+		if ( r_shadowMapSinglePass.GetBool() ) {
+			void RB_ShadowMap_RenderAllLights();
```
stgatilov: I know this is not your code, but... extern it, please! =)

stgatilov (commented):
Oh my god, I have finished it!!!

renderer/backend/GpuBuffer.cpp
stgatilov: It's magic? =)

    That is, data written to the store by either the client or server will
    be visible to any subsequently issued GL commands with no further action
    taken by the application.

It seems that `visible to any subsequently issued GL commands` has replaced the initial wording `immediately visible` (Public Bug 935 in specs).

I see two ways it can be implemented:

 * **API way**. At the closest GL API call, the driver copies the whole mapped range. This is very bad because all three buffers are mapped, meaning that implementation can copy all of them every frame.

 * **Paging way**. The mapped memory pages are specifically marked, so that first write to them triggers page fault or similar exception. Driver catches the exception and marks the page as modified somewhere. On the first GL call it uploads all modified pages. This is pretty strange understanding, because it would mean cross-platform OpenGL standard requires a memory-mapping unit on host (maybe it is OK today). However, it also means that memory is uploaded in pages with at least 4KB alignment, so we should force-align `frameSize` to page size, because otherwise we can accidentally upload a bit of data which we can potentially be drawn by GPU right now.

renderer/backend/DepthStage.cpp
fholger: Yes, absolutely. I honestly don't know why I didn't do it immediately, but no one should have to read (or write) that abomination of a check more than once :D

renderer/backend/GpuBuffer.cpp
stgatilov: Yes, I think something like inline `GpuBuffer::AlignSize` method would be enough.

renderer/backend/GpuBuffer.cpp
stgatilov: Yes, "indexed target" is also good clarification.

renderer/backend/GpuBuffer.cpp
stgatilov: Moreover, there seems to be no way to learn of the page size, if paging is used in implementation. Which is even more weird. As for potential overwrite, perhaps driver simply allows it because logically we must not change the actual data with it.

fholger: Again, no idea how they do it, but I'm reasonably certain the usage pattern I employ is correct and performant. It is essentially as proposed in the "approaching zero driver overhead" talks by the people who actually work on the GPU drivers, so I'd hope they know what they're doing :D 
But I do think the mapped buffer memory on the CPU side is some kind of specially mapped construct controlled by the driver, so I'd speculate that your "paging way" may touch on the actual implementation. But I think the control/tracking would have to be as fine as the minimum buffer alignment requirements, which I think would be 32 bytes, to guarantee correctness.

renderer/backend/GpuBuffer.cpp
fholger: Either way, I believe it should not require significant refactoring of the `GpuBuffer`, so I'd like to postpone this to a later PR when we might actually refactor `VertexCache` to use this class :)

renderer/backend/StencilShadowStage.cpp
fholger: I think that might be incorrect, since even with r_shadows == 2, some lights fall back to stencil shadowing.

renderer/backend/DepthStage.cpp
fholger: The name should be `depth`, but that's unfortunately already the name of the depth shader in the "old" backend, which I tried not to touch at all. I guess it doesn't hurt to rename the files for now, but perhaps I'll add a TODO to rename shader and files to depth once the old code is gone :)

renderer/backend/DepthStage.cpp
fholger: Yeah... my feeling is that some things could potentially be simplified, but it's not so easy in part due to the material system that basically predates the programmable pipeline and shaders and was only retrofitted to work with them :) It's even worse in the interaction stage.

renderer/backend/DepthStage.cpp
fholger: Actually, I was wrong. The existing shader is called `depthAlpha`, so there's no reason for the new one not to be named `depth` :)

renderer/backend/StencilShadowStage.cpp
fholger: Actually, I already added a specific function to the vertexCache, because using `VertexPosition` like this is a terrible hack. Just forgot to use it here :)

renderer/backend/InteractionStage.cpp
fholger: No, I think the maximum at any point is 7 - normal, diffuse, specular for the objects, falloff and projection for the light and depth and stencil for soft stencil shadows :)

renderer/backend/InteractionStage.cpp
fholger: Yes, good point. I'm not 100% certain, but since the buffer technically gets bound to the shader state even if soft shadows are deactivated, I wanted to play it safe to avoid potential issues that only affect e.g. AMD cards and that I thus wouldn't notice. But you're right in that it certainly doesn't need to be initialized to its max size...

renderer/backend/InteractionStage.cpp
fholger: What kind of overflow? `glBufferData` (not SubData) reinitializes the buffer, so there's no danger to write beyond the buffer's reach.

renderer/backend/StencilShadowStage.cpp
fholger: In general I find it cleaner to not make something a member if it doesn't have to be. Avoiding the allocations feels like a micro-optimization that I would expect won't make a measurable difference, but I'll give it a go just to be safe.

renderer/backend/DrawBatchExecutor.cpp
fholger: For now, I've reverted it to always count the number of draw surfaces, which is consistent with what it's been doing so far. Adding another counter is a bit more effort, and it's probably irrelevant for most people. If you really need to see/measure the actual draw calls issued, tools like RenderDoc or nSight will give you a better insight, anyway.

renderer/backend/DrawBatchExecutor.h
fholger: I reworded the explanation, is it better?

renderer/backend/InteractionStage.cpp
fholger: Does not make a difference that I could measure, so I'd leave it here. But I've changed it to idList.

renderer/backend/StencilShadowStage.cpp
fholger: Makes no discernible difference, so I'd leave them as local variables. But I changed them to `idList`.

fholger:
> Oh my god, I have finished it!!!

@stgatilov Thanks for doing this! I know reviewing PRs of this size is a pain, and I certainly hope it will not become the norm. Even so and despite our disagreements on some of the finer details, I think this was well worth it. It was absolutely vital to get a second pair of eyes on this, and I think overall the code has benefitted from your input :)

renderer/backend/GpuBuffer.cpp
stgatilov: This can't be "memory controlled by driver", this is memory directly accessed by CPU instructions. It has to use some mechanism supported by CPU.

I think a more likely possibility is using write-combined memory for mapped regions. This memory is uncached, but subsequent writes to such memory are hold in WC buffer in order to combine them to a larger write. If this memory region points directly to GPU memory, then it will make changes visible sort-of immediately. Maybe some flush is necessary to push the last bits, but it should be cheap to do at any API call. However, if there are separate memory buffers on CPU and on GPU, then it is unclear how could driver understand when to copy.

renderer/backend/InteractionStage.cpp
stgatilov: I mean the case when user sets `r_softShadowQuality 200` and gets a crash =)

Maybe just get the cap from shader, then check for overflow, and reduce the cvar if it happens...

renderer/backend/DrawBatchExecutor.cpp
stgatilov: I think many mappers look at draw calls count using r_showPrimitives or somewthing like that. I guess they would be disappointed if you radically change this metric in a way like you did =)

Regarding new counter, I think the question is: can it be useful on any other machine except for developers' machines? Can you see the number easily in RenderDoc?

renderer/backend/GpuBuffer.cpp
fholger: Well, you clearly understand more of this than I do. Unfortunately, I didn't find any deeper technical explanation about how that COHERENT_BIT part is actually implemented.

My understanding of the techniques used in this PR mostly come from these two talks:
https://www.youtube.com/watch?v=K70QbvzB6II (AZDO: approaching zero driver overhead) and
https://www.youtube.com/watch?v=-bCeNzgiJ8I (Beyond Porting)
If you don't know them, you might find them interesting for insight into how persistent buffer, multidraw and bindless textures work together to eliminate driver overhead. But unfortunately neither of them goes into detail of how persistent buffers actually work :-/

renderer/backend/DrawBatchExecutor.cpp
fholger: I'm not sure you can see the number easily in RenderDoc. In nSight, you can profile a frame, which will tell you the number of issued actual draw calls.

But I think the entire point of this is that the number of draw calls should be reduced to the point that the exact number of either the draw calls or the underlying drawn objects should not matter too much anymore, so mappers can concentrate on the number of tris and vertices, instead. So I'm not sure the extra counter is all that useful. So I'd suggest we add it if someone does find a use for it :)

renderer/backend/GpuBuffer.cpp
fholger: This post suggests that persistent buffers may reside in a shared memory region accessible to both CPU and GPU, so there may not actually be any memory transfer: https://community.khronos.org/t/persistent-buffer-mapping-what-kind-of-magic-is-this/71646

stgatilov: Yes, looks like the memory is not duplicated, so the version of non-cacheable WC writes is the most likely. They get flushed with any CPU-serializing instruction like SFENCE or CPUID. It's quite likely that most of API calls do it with low performance cost.

PR #21, opened 2020-06-13
Frontend parallelization

fholger:


idlib/ParallelJobList.cpp
```diff
@@ -25,8 +25,8 @@ If you have questions concerning this license or the applicable additional terms
 
 ===========================================================================
 */
-#pragma hdrstop
 #include "precompiled.h"
+#pragma hdrstop
```
stgatilov: I wonder if these hdrstop-s were taken from Doom 3 BFG source code.

I can hardly understand what they do, given that we include precompiled header explicitly. Except for a chance to have an error by misplacing it.

renderer/Interaction.cpp
```diff
@@ -30,6 +31,60 @@ idInteraction implementation
 idCVar r_useInteractionTriCulling("r_useInteractionTriCulling", "1", CVAR_RENDERER | CVAR_BOOL, "1 = cull interactions tris");
 idCVarInt r_singleShadowEntity( "r_singleShadowEntity", "-1", CVAR_RENDERER, "suppress all but one shadowing entity" );
 
+void idInteraction::PrepareLightSurf( drawSurf_s **link, const srfTriangles_t *tri, const viewEntity_s *space,
```
stgatilov: Is it right that `R_LinkLightSurf` is going to be deleted?
And you don't do it now to avoid potential conflicts?...

renderer/tr_light.cpp
```diff
@@ -14,14 +14,13 @@
...
-#include <set>
-#include <unordered_set>
 #pragma hdrstop
 
 #include "../idlib/geometry/sys_intrinsics.h"
 #include "tr_local.h"
 #include "Model_local.h" // Added in #3878 (soft particles) to allow r_AddAmbientDrawSurfs to access info about particles to 
-						 // pass to the backend without bloating the modelSurface_t struct used everywhere. That struct is the only
+#include "Profiling.h"
+// pass to the backend without bloating the modelSurface_t struct used everywhere. That struct is the only
```
stgatilov: Comment split?
I think either put this comment on separate lines before `#include "Model_local.h"` or just delete it.

renderer/tr_light.cpp
```diff
@@ -1548,10 +1565,10 @@ void R_AddSingleModel( viewEntity_t* vEntity ) {
 	if ( !vEntity->scissorRect.IsEmpty() || R_HasVisibleShadows( vEntity ) ) {
 		model = R_EntityDefDynamicModel( &def );
 		if ( model == NULL || model->NumSurfaces() <= 0 ) {
-			if ( def.parms.timeGroup ) {
+			/*if ( def.parms.timeGroup ) {
```
stgatilov, line 1563: I guess it was used for slowing time around player for [The Artefact](https://doom.fandom.com/wiki/The_Artifact).
It is not used, so probably safe to disable.

renderer/Interaction.cpp
```diff
@@ -1130,6 +1187,15 @@ bool idInteraction::IsPotentiallyVisible( idScreenRect &shadowScissor ) {
 	return true;
 }
 
+void idInteraction::LinkPreparedSurfaces() {
+	for (auto it : surfsToLink) {
+		// actually link it in
+		it.surf->nextOnLight = *it.link;
+		*it.link = it.surf;
+	}
+	surfsToLink.Clear();
```
stgatilov: `Clear` frees resources. Is it indended?
To be honest, allocating buffer for every interaction every frame sounds like a performance problem.

Perhaps the best solution would be to have storage for a few elements embedded into `idInteraction`, and switch to heap buffer only when this capacity is exceeded. Unfortunately, I don't see such container in idLib... that's the mindset of "we can always put hard cap on any array" =(

The simpler solution would be maintaining one large array instead of per-interaction arrays, but that would need synchronization. Although malloc calls are also working with single resource and thus doing synchronization, although with atomics only.

renderer/tr_light.cpp
```diff
@@ -1007,6 +1008,7 @@ it and any necessary overlays
 ===================
 */
 idRenderModel *R_EntityDefDynamicModel( idRenderEntityLocal *def ) {
+	std::lock_guard<std::mutex> lock (def->mutex);
```
stgatilov: C++11 threading spreads over the code...

Meahwhile, there are two in-house libraries for multithreading. The one from original Doom 3 is pretty limited (4 numbered critical sections), but the one which came from Doom 3 BFG supports creating arbitrary number of mutexes.

Why not use `idSysMutex` and `idScopedCriticalSection` then?

renderer/tr_light.cpp
```diff
@@ -1156,31 +1158,32 @@ void R_AddDrawSurf( const srfTriangles_t *tri, const viewEntity_t *space, const
...
-	Sys_EnterCriticalSection( CRITICAL_SECTION_TWO );
-	if ( tr.viewDef->numDrawSurfs == tr.viewDef->maxDrawSurfs ) {
-		drawSurf_t	**old = tr.viewDef->drawSurfs;
-		int			count;
+	if (!deferred) {
+		// bumping this offset each time causes surfaces with equal sort orders to still
+		// deterministically draw in the order they are added
+		tr.sortOffset += 0.000001f;
+		// if it doesn't fit, resize the list
+		Sys_EnterCriticalSection( CRITICAL_SECTION_TWO );
```
stgatilov: This was added by Duzenko when he worked on the same task. So I think it would be right to replace it with object mutex, like `idSysMutex`.

renderer/tr_light.cpp
```diff
@@ -1156,31 +1158,32 @@ void R_AddDrawSurf( const srfTriangles_t *tri, const viewEntity_t *space, const
...
-	// bumping this offset each time causes surfaces with equal sort orders to still
-	// deterministically draw in the order they are added
-	tr.sortOffset += 0.000001f;
-
-	// if it doesn't fit, resize the list
-	Sys_EnterCriticalSection( CRITICAL_SECTION_TWO );
-	if ( tr.viewDef->numDrawSurfs == tr.viewDef->maxDrawSurfs ) {
-		drawSurf_t	**old = tr.viewDef->drawSurfs;
-		int			count;
+	if (!deferred) {
```
stgatilov, line 1181: Is there any place which uses non-deferred case?
If yes, than can't it be reworked to also use deferred code path?

Also, right now you have this code in two places, which is a bad idea.

renderer/tr_light.cpp
fholger: Wouldn't it be better to use the std::mutex and get rid of idSysMutex, instead? Not only is it less code to maintain for us, but idSysMutex uses a CRITICAL_SECTION on Windows, and since VS 2017 (or even 2015?) the std::mutex implementation uses SRWLock, which should be faster.

renderer/Interaction.cpp
fholger: At the moment, it's still used in `R_AddLightSurfaces`, which has not been parallelized so far. We might do that in the future (it is the next biggest "offender" in the frontend, although with some distance to `R_AddModelSurfaces`), but that should probably be its own PR :)

So yeah, I guess I didn't want to touch `R_AddLightSurfaces` in this PR, hence why the function is still around.

idlib/ParallelJobList.cpp
fholger: You're right, that's copy&paste... I'm also not entirely clear on its purpose. The MS documentation speaks of saving compiler state up to this point, but doesn't really convey the impact of using it. Guess I'll just remove it.

renderer/tr_light.cpp
fholger: Yes, unfortunately `R_AddDrawSurf` is used in several places, and I'm not familiar with all of them. So I opted for the `deferred` parameter to not have to touch all of those places for now. But yeah, I should probably extract the list resize into its own function.

renderer/tr_light.cpp
fholger: I'll need to check, but there's a good chance that the non-deferred code path never executes in parallel, anyway, so might just be able to remove the synchronization entirely.

renderer/tr_light.cpp
fholger: Ah, interesting.
Yeah, I combed our code base and am reasonably certain that it's not used and should probably be deleted - if we ever do need it, it would probably require some significant overhaul, not just to make it thread-safe.

renderer/Interaction.cpp
fholger: No, not intentional :( I expected std::vector behaviour which just removes the elements from the list, not delete the underlying storage...

I guess I could opt for the manual linked list, since I can reuse the existing `nextOnLight` pointer. Not an elegant solution, either, but at least it's an established pattern. Let's see if it makes any difference in performance.

renderer/tr_light.cpp
fholger: Ah, yes, Resharper automatically added the `#include`, I guess it wasn't smart enough to recognize the comment block :D

I'll just delete it.

renderer/tr_light.cpp
fholger: Ok, after looking at it again, it turns out 3 of 4 uses of `R_AddDrawSurf` resulted from recursive calls from within `R_AddDrawSurf`, so those were a no-brainer to switch to deferred. The only remaining usage is within GUI code, of which parts are also a recursive calls, but parts are not. And since GUI itself needs to be protected if run in parallel, this is tricky to move to the deferred path. So I'm leaving that for now, but I'm now also fairly sure that the remaining use of the non-deferred path is not run in parallel, so no synchronization is needed.

renderer/Interaction.cpp
fholger: Well, I exchanged it for linked lists, which makes the interface and code a tad uglier, but avoids the allocation issue. However, far as I can tell, it makes no measurable difference in-game. Let me know what you think.

renderer/tr_light.cpp
stgatilov: I have a feeling that it's better to change underlying implementation of `idSysMutex`, if there is something faster than CriticalSection. I tried to find something about the performance difference, and the first argument was: SRWLock is 8 bytes instead of 40 bytes of critical section. Ha-ha, std::mutex is 48 bytes =)

By the way, SRW sounds like reader-writer lock. Shouldn't it be something that works good when there are many read-only users? If yes, then it needs different interface, I guess.

renderer/Interaction.cpp
stgatilov: Well `idList::SetNum(0)` does not free memory. Linked lists are OK, I guess --- they will traversed only once, and won't contain much data.

renderer/Interaction.cpp
stgatilov: I think that function needs to be removed. If not here, then by the other change soon. It contains some logic which is dangerous to have in two places.

renderer/tr_light.cpp
stgatilov: Oh, std::mutex is 80 bytes in 64-bit mode =)

fholger: But why is it better? It would mean additional coding effort, and again, code that we'd have to maintain and test across our platforms, instead of a solution that we get out of the box. And sure, std::mutex has the usual bloat of STL, but if we ever use so many mutexes that that's going to be an issue, we'll have other problems :)

SRWLocks, according to their documentation, have two modes of operation. One is indeed many readers/one writer, the other is a single exclusive access. So I guess MS crammed two different concepts into the same structure :D

renderer/Interaction.cpp
fholger: Fair point, I've removed the code duplication, now only the linking part takes a different turn depending on the caller. (Which it has to, anyway, because the usage in R_AddLightSurfs is a bit different.)

renderer/tr_light.cpp
stgatilov: No coding effort is necessary. Just replace `std::mutex` with `idSysMutex`, and everything will continue working. The performance difference is hypothetical both ways =)

Mixing different frameworks does not sound well. BFG code already uses idSys primitives. I have already used a few to make idFileSystem thread-safe.

One nasty question I had recently is compatibility of C++11 with pthreads. Some time ago we had problems with Linux build hanging or crashing on exit. At that moment I tried to search for an answer to this simple question, but surprisingly did not find any.

And on the point of bragging, me and Duzenko had bad experience with `std::condition_variable`. In my opinion it is too low-level to be used by itself and too easy to misuse. Win32 events make more sense, which we have in form of idSysSignal.

fholger: Fine, I've changed the two mutexes to idSysMutex.

But honestly, I think your disdain for the STL and simultaneous reverance of id code sometimes borders on religious ;)
The only reason we even have idSysMutex is because duzenko copied thousands of lines of code from BFG, which is code that is completely untested outside of its BFG habitat, and I've already had to modify it to sensibly support more than 2 worker threads. There were C++ task libraries (building on top of C++14 threading primitives) available that could have done the same job for us without now having to be responsible for that BFG threading code, and I think they would have been the smarter choice. If we ever find the need to establish a new platform (or even reactivate macOS, although with OpenGL being deprecated, I guess we'll have avoided that one), this notion of relying on id platform code is going to bite us in the ass...

stgatilov: Thank you!

I'm not exactly "fond" of the ID code, to be honest. The `idList` container is pretty weird, bears too little resemblance to STL to be intuitive, and its original code even lacked some things like exponentially growing push_back. But we have it, and it works.

As for STL... yes, I dislike it a lot. Call it religion if you like. The previous TDM lead seemed to have boost religion BTW =) We can start a holywar on the topic and have a lot of fun =)

One major point against STL is debug performance. It just sucks, for no apparent reason. And just like build time, debug performance does not drop dramatically every time you use STL. It goes down gradually and without notice. No particular commit can be blamed for it, the effect of one small change can't be measured. But in the end STL is everywhere, and debug performance sucks, and everyone is sure this is normal. On my daily job I don't care much, because I can record the calls to my code, and replay it without any external code. But for TDM debug performance is very important.

I have recently wrote a [small article about debug performance](https://dirtyhandscoding.github.io/posts/fast-debug-in-visual-c.html) (it even mentions TDM). And I guess you won't be surprised if I say that I am writing a replacement for some bits of STL (don't worry, it won't get into TDM). And its vector works much faster than the one from STL in debug. E.g. if I only enable \Ob1, it is 5 times faster on the benchmark from the article. It is the inherent cost of STL, it will never go away.

Anecdotical cases include the story about you using `std::regex` in shader manager. Its debug performance is so bad that it really hurts!
The sad fact is that when files are packaged into pk4 files, the TDM packager uses regexes to decide what to include and what not, and which pk4 to put it into. Packaging TDM in debug takes hours! I am happy that I don't need to debug it often.

fholger:
Merged to SVN trunk.

PR #23, opened 2020-07-27
Gamepad support prototype

fholger:


ThirdParty/conanfile.py
```diff
@@ -55,6 +57,7 @@ class TdmDepends(ConanFile):
         "vorbis/1.3.6@bincrafters/stable",
         "fltk/1.3.5@thedarkmod/local",
         "BLAKE2/master@thedarkmod/local",
+        "glfw/3.3.2@",
```
stgatilov, line 60: What does it mean?....

I know new conan allows not setting user/channel, but I thought there should be no @ then =)

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+		int action;
+	};
+
+	enum padStatus_t {
+		STATUS_NONE,
+		STATUS_PRESSED,
+		STATUS_RELEASED,
+	};
+
+	struct padState_t {
```
stgatilov: Sounds like the state of the whole gamepad.
But it seems to only the state of single "button", no?...
Maybe `padButtonState_t`?

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+		int activeBinding;
+
+		padState_t() : status(STATUS_NONE), lastPressed(0), lastReleased(0), activeBinding(-1) {
+			for ( int i = 0; i < BIND_NUM; ++i ) {
+				bindings[i].action = -1;
+			}
+		}
+	};
+
+	padState_t *buttonState = nullptr;
```
stgatilov: Can we make this a static array?

I guess having it dynamic is pointless, and seeing array is more obvious.

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	int button = idGamepadInput::StringToButton( args.Argv( 2 ) );
+	if ( button == -1 || button >= PAD_NUM_BUTTONS ) {
+		common->Printf( "\"%s\" isn't a valid gamepad button\n", args.Argv(2) );
+		return;
+	}
+
+	// copy the rest of the command line
+	for ( int i = 3; i < numArgs; i++ ) {
```
stgatilov, line 134: (cosmetic) probably move `cmd` declatation here?...

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+			cmd += " ";
+		}
+	}
+
+	idGamepadInput::SetBinding( button, type, cmd );
+}
+
+void ArgCompletion_PadButtonName( const idCmdArgs &args, void(*callback)( const char *s ) ) {
+	for ( auto kn = buttonNames; kn->name; kn++ ) {
+		callback( va( "%s %s", args.Argv( 0 ), kn->name ) );
```
stgatilov, line 146: Recently I added arg completion too. 
Still don't understand wtf `args.Argv(0)` is doing here =))

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+	for ( nameMapping_t *n = buttonNames; n->name; n++ ) {
+		if ( !idStr::Icmp( str, n->name ) ) {
+			return n->value;
+		}
+	}
+
+	return -1;
+}
+
+int idGamepadInput::StringToType( const idStr &str ) {
```
stgatilov, line 172: I think these methods should be one, receiving `nameMapping *enumTable` as argument.

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+	enum padStatus_t {
+		STATUS_NONE,
+		STATUS_PRESSED,
+		STATUS_RELEASED,
+	};
+
+	struct padState_t {
+		padStatus_t status;
+		uint64_t lastPressed;
+		uint64_t lastReleased;
```
stgatilov, line 48: I would appreciate if you write at least here what sort of time is stored here, e.g. astronomical or game time, in which units.

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+		buttonState[i].status = STATUS_NONE;
+		buttonState[i].lastPressed = 0;
+		buttonState[i].lastReleased = 0;
+		buttonState[i].activeBinding = -1;
+	}
+
+	if ( type == BIND_MODIFIER ) {
+		// setting the modifier key unbinds all other actions for that button
+		for ( int i = 0; i < BIND_NUM; ++i ) {
+			buttonState[button].bindings[type].action = -1;
```
stgatilov, line 200: Why for i?
Maybe here should be `i` instead of `type` ?

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+void idGamepadInput::SetButtonState( int button, bool pressed ) {
+	if ( button < 0 || button >= PAD_NUM_BUTTONS ) {
+		return;
+	}
+
+	padState_t &st = buttonState[button];
+
+	if ( pressed ) {
```
stgatilov, line 225: Question: if button is already pressed, should we really update time?

framework/GamepadInput.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+	const uint64_t longPressTime = in_padLongPressTimeMs.GetInteger() * 1000;
+	const uint64_t twoButtonTime = in_padTwoButtonMaxTimeMs.GetInteger() * 1000;
+
+	const uint64_t currentTime = Sys_GetTimeMicroseconds();
+
+	// double-lean hack
+	if ( !forwardLeanActive ) {
+		if ( leanLeftPressed > 0 && leanRightPressed > 0 ) {
+			common->Printf("Forward lean activated.\n");
```
stgatilov: This looks like console spam.

Honestly, if you need this as debug output, you should probably use `DM_LOG`. You can even create a channel for this.

framework/Session_menu.cpp
```diff
@@ -277,6 +277,15 @@ idSessionLocal::SetPbMenuGuiVars
 void idSessionLocal::SetPbMenuGuiVars( void ) {
 }
 
+void idSessionLocal::ToggleMenu() {
+	if ( guiActive != guiMainMenu ) {
+		console->Close();
+		StartMenu();
+	} else if ( mapSpawned ) {
+		ExitMenu();
+	}
```
stgatilov: Why can't you use the same way for switching to main menu as key input uses?
Everything related to main menu and switching GUIs is rather complicated, and I would prefer to avoid having two pieces of code which does it.

I think bringing up main menu is done at the beginning of `idSessionLocal::ProcessEvent`. Exiting the menu is handled in `onEsc` handler in `mainmenu_utils.gui`.

Maybe it would be better if you bind something on game pad to virtual "escape", and then pass it to GUI, and handle for bringing menu? Maybe it would even easier if you simply redirect this gamepad's virtual escape to keyboard escape, so that it starts working as escape everywhere?

framework/UsercmdGen.cpp
```diff
@@ -667,6 +709,8 @@ void idUsercmdGenLocal::MouseMove( void ) {
 idUsercmdGenLocal::JoystickMove
 =================
 */
+idCVar in_invertYawAxis( "in_invertYawAxis", "0", CVAR_GAME|CVAR_BOOL|CVAR_ARCHIVE, "Invert gamepad yaw axis" );
+idCVar in_invertPitchAxis( "in_invertPitchAxis", "0", CVAR_GAME|CVAR_BOOL|CVAR_ARCHIVE, "Invert gamepad pitch axis" );
```
stgatilov: Probably add "pad" here too.

I suppose there may be similar cvars for mouse too.

framework/UsercmdGen.cpp
```diff
@@ -676,15 +720,12 @@ void idUsercmdGenLocal::JoystickMove( void ) {
...
-		cmd.forwardmove = idMath::ClampChar( cmd.forwardmove + joystickAxis[AXIS_FORWARD] );
-	}
-
-	cmd.upmove = idMath::ClampChar( cmd.upmove + joystickAxis[AXIS_UP] );
+	float yawAngleSpeed = anglespeed * ( in_invertYawAxis.GetBool() ? 1 : -1 );
+	float pitchAngleSpeed = anglespeed * ( in_invertPitchAxis.GetBool() ? 1 : -1 );
+	viewangles[YAW] += yawAngleSpeed * in_yawSpeed.GetFloat() * joystickAxis[AXIS_YAW] / 127.f;
+	viewangles[PITCH] += pitchAngleSpeed * in_pitchSpeed.GetFloat() * joystickAxis[AXIS_PITCH] / 127.f;
+	cmd.rightmove = idMath::ClampChar( cmd.rightmove + joystickAxis[AXIS_SIDE] );
+	cmd.forwardmove = idMath::ClampChar( cmd.forwardmove + joystickAxis[AXIS_FORWARD] );
```
stgatilov, line 728: Does it mean that joystick actual position specifies "acceleration" of movement, not "velocity"?
Or it means they are added to mouse input?...

game/Player.cpp
```diff
@@ -7752,6 +7756,15 @@ void idPlayer::RouteGuiMouse( idUserInterface *gui ) {
 		oldMouseX = usercmd.mx;
 		oldMouseY = usercmd.my;
 	}
+	if ( usercmd.jx != 0 || usercmd.jy != 0 ) {
+		extern idCVar in_padMouseSpeed;
+		float speed = in_padMouseSpeed.GetFloat() / 640.f * renderSystem->GetScreenWidth();
+		// hack: we want the original axis deflection, without inversed axes if enabled
+		int x, y;
+		Sys_GetCombinedAxisDeflection( x, y );
```
stgatilov, line 7764: Question: why you added jx/jy members to usercmd_t instead of converting this data into mouse movements and writing it to mx/my?

ThirdParty/conanfile.py
fholger: The package is in the main conan channel, and strangely, this is the only working way to address it. If you omit the @, it errors out. If you add conan/stable, it doesn't find it either...

framework/GamepadInput.cpp
fholger: I admit I just copied that code from some other place. From my understanding, the code completion expects the entirety of the typed command, and using args.Argv(0) is probably just a simple way to pass through the original spelling (upper/lower casing) that the user used for the command itself without modifying it in the process?

framework/GamepadInput.cpp
fholger: No, that was just a temporary debug output that I forgot to remove :)

framework/Session_menu.cpp
fholger: This is what I originally did, hardcoding the gamepad button to K_ESCAPE, but it's pretty terrible and I want to avoid it.

To start with, imho hardcoding the ESCAPE key is bad and a hack in the first place (completely circumventing the otherwise customizable input system), so I don't want to further cement this as I'd rather see it refactored some time in the future. Secondly, emitting a K_ESCAPE event for a gamepad button is another hack on top of that which creates problems for bindability of the gamepad button in question. In fact, my original solution just hardcoded the START button to K_ESCAPE, which also prevented it from binding long press or modifier actions etc.. So either way, I need a command that the button can be bound to, and imho having a command `emulateEscape` is just weird.

If you are concerned about code duplication, I'd rather change the existing handling of the ESC key in the GUI context to the new command than the other way around.

framework/UsercmdGen.cpp
fholger: Both are reset for every issued cmd, so it's not acceleration. But they may be preset with keyboard input.

This isn't something any user would reasonably trigger in real life, obviously, but if not using + you'd have to ensure that these lines only fire if there's actual stick movement, so that's a potential for bugs.

game/Player.cpp
fholger: Because `mx` and `my` are not movements, but absolute mouse positions and therefore need to be handled quite a bit differently from joystick input. I originally wanted to do what you suggest, but it's simply not possible. At least not without some refactoring, and that seemed risky...

framework/UsercmdGen.cpp
fholger: Yep, no idea why I didn't add it here. Probably an oversight.

ThirdParty/conanfile.py
stgatilov: I don't see a package with @. Moreover, I used some packages which look this way, and I did not put @.
As for conan/stable, that's the wrong channel. It is in bincrafters/stable.

image

framework/GamepadInput.cpp
stgatilov: The only explanation which comes to my mind is that they wanted to allow changing the command which you auto-complete. Like you enter "hit ", then press Tab, and it turns into "bash hit", or something like that =)

framework/Session_menu.cpp
stgatilov: Yes, my suggestion is to turn it into "escape" key after all the binds. Instead of binding something to "togglemenu", bind it to "escapekey" or "seEscape", which generates escape key press.

I'm not sure there is anything to refactor. `sysEvent_t` sounds like a system event, not raw key press. The fact that it reuses key codes for Escape, functional keys, and maybe something else is probably 1) for convenience, 2) because there was no other input device supporting that many buttons anyway.
With such thinking, generating a sysEvent_t with escape is not a hack really.

ThirdParty/conanfile.py
fholger: No, the bincrafters one is explicitly obsolete, see here: https://github.com/bincrafters/conan-glfw - it's been moved to the conan-center index.
The correct one is the second-to-last in that list, without any channel. However, if I put it like that in the file, it is not found. I needed to add the '@' at the end to make it work. I don't know why, it's just the way it is.

framework/Session_menu.cpp
fholger: I disagree. `sysEvent_t` is just a vehicle to transport certain events around the engine, and it has a specific `type` field that can be `SE_KEY`, `SE_MOUSE` or other things. The check is specifically for an `SE_KEY` event of type `K_ESCAPE`, and only keyboard input devices would ever generate such an event. This is not convenience, this is a hardcoded key mapping.
Worse, it's used in several places in UI code for different things that don't really make sense to have on a gamepad bind. While thankfully not really harmful, it just muddies the water.
Finally, the method to generate any such `sysEvent_t` event was actually internal to the `Sys_` platform functions and not publicly exposed before this PR. So injecting this kind of event externally is definitely a hack.

But whatever. I've changed it as you requested. After all, why would we want to clean up UI interactions?

stgatilov: It was possible to generate sysEvent_t before, like this:
```
void Automation::PushInputEvents(int key, int value) {
	sysEvent_t ev;
	memset(&ev, 0, sizeof(ev));
	ev.evType = SE_KEY;
	ev.evValue = key;
	ev.evValue2 = value;
	eventLoop->PushEvent(&ev);
}
```
I think it is not the only such key. I'm pretty sure arrows, delete, and backspace in GUI code are also hardcoded to K_XXX constants.

framework/UsercmdGen.cpp
stgatilov: Ok, I see

game/Player.cpp
stgatilov: This is a bit sad.
Thankfully there are not too much code depending on these coordinates, so why not =)

ThirdParty/conanfile.py
stgatilov: I think I will look into it after merge...

game/Player.cpp
fholger: Yeah... I've come to the conclusion that the UI code is really not a highlight of the engine. It's some of the less appealing code I've seen so far. A lot of it seems hacked in and not easily extensible. I think I can understand why they replaced it for the BFG version...

PR #24, opened 2020-08-22
Remove old GL stuff

fholger:


renderer/Material.cpp
```diff
@@ -1562,7 +1532,6 @@ void idMaterial::ParseStage( idLexer &src, const textureRepeat_t trpDefault ) {
 					newStage.vertexProgram = R_FindARBProgram( GL_VERTEX_PROGRAM_ARB, token.c_str() );
```
stgatilov: I think the code for vertex program should be removed too.

renderer/Material.h
```diff
@@ -175,17 +175,14 @@ static const int	MAX_FRAGMENT_IMAGES = 8;
...
 	int					vertexParms[MAX_VERTEX_PARMS][4];	// evaluated register indexes
 
-	int					fragmentProgram;
 	int					numFragmentProgramImages;
 	idImage *			fragmentProgramImages[MAX_FRAGMENT_IMAGES];
 
 	//idMegaTexture		*megaTexture;		// handles all the binding and parameter setting 
 
-	bool				GLSL;
 	//stgatilov: trying to use new shader framework
```
stgatilov: Comment worth changing? =)

renderer/VertexCache.cpp
```diff
@@ -334,13 +325,8 @@ void idVertexCache::EndFrame() {
 
 	// ensure no GL draws are still active on the next buffer to write to
 	int nextListNum = ( listNum + 1 ) % VERTCACHE_NUM_FRAMES;
-	if ( r_useFenceSync.GetBool() ) {
-		LockGeoBufferSet( backendListNum );
-		WaitForGeoBufferSet( nextListNum );
-	} else {
-		// this is going to stall, but it probably doesn't matter on such old GPUs...
-		qglFinish();
```
stgatilov: To be honest, it's probably worth leaving `r_useFenceSync` in the code. I remember that many players (especially with old hardware) used it to cover some issues they had. If we remove it, they will probably get stuck with their problem.

fholger: Hm, I wonder. The only actual report I can clearly remember was for an old Intel card that was stuck on GL 3.1 plus extensions, so that card is out going forward, anyway. The remaining reports are less clear, because unfortunately someone usually posts a collection of possible cvars including this particular one, and it then gets rather difficult to track what actually helped.

Given that fence syncs are core since GL 3.2 and such a vital feature, I'd rather like to get rid of the switch and see if there are really any cards left that claim to implement GL 3.3, but have issues with this. Perhaps we could then even figure out what is actually the issue and perhaps fix that or work around it instead of just falling back on the awful glFinish.

renderer/Material.h
fholger: Worth removing, anyway :)

renderer/Material.cpp
fholger: I just deleted that whole #ifdef block. Wouldn't have done much good anymore, anyway.

PR #26, opened 2021-04-03
Improve level load times

fholger:


idlib/ParallelJobList.cpp
```diff
@@ -126,7 +126,7 @@ idParallelJobList_Threads
 ================================================================================================
 */
 
-static idCVar jobs_longJobMicroSec( "jobs_longJobMicroSec", "10000", CVAR_INTEGER, "print a warning for jobs that take more than this number of microseconds" );
+static idCVar jobs_longJobMicroSec( "jobs_longJobMicroSec", "500000", CVAR_INTEGER, "print a warning for jobs that take more than this number of microseconds" );
```
stgatilov: Isn't it a problem for typical jobs which run within frame render?
You are efficiently silencing the warning.

renderer/Image_load.cpp
```diff
@@ -552,11 +552,18 @@ void idImage::GenerateImage( const byte *pic, int width, int height,
 	GL_CheckErrors();
 	GL_SetDebugLabel( GL_TEXTURE, texnum, imgName );
 
-		//Routine test( &uploading );
-		auto start = Sys_Milliseconds();
+	//Routine test( &uploading );
+	auto start = Sys_Milliseconds();
+	if ( GLAD_GL_ARB_texture_storage && !generatorFunction ) {
+		int levels = 1 + idMath::Floor( idMath::ILog2( Max( scaled_width, scaled_height ) ) );
```
stgatilov: If Floor necessary?
ILog2 is already integer-valued logarithm --- it only takes exponent from float representation.

renderer/Image_load.cpp
```diff
@@ -552,11 +552,18 @@ void idImage::GenerateImage( const byte *pic, int width, int height,
 	GL_CheckErrors();
 	GL_SetDebugLabel( GL_TEXTURE, texnum, imgName );
 
-		//Routine test( &uploading );
-		auto start = Sys_Milliseconds();
+	//Routine test( &uploading );
+	auto start = Sys_Milliseconds();
+	if ( GLAD_GL_ARB_texture_storage && !generatorFunction ) {
```
stgatilov: An additional troubleshooting cvar would be helpful.
So that we could help a poor guy with an old AMD gpu with buggy drivers =)

stgatilov: What's the problem with `generatorFunction` here?

renderer/qgl.h
```diff
@@ -24,6 +24,7 @@
 #define GL_KHR_debug							1	//core since 4.3
 #define GL_ARB_stencil_texturing				1	//core since 4.3
 #define GL_ARB_buffer_storage					1	//core since 4.4
+#define GL_ARB_texture_storage					1	//core since 4.2
```
stgatilov, line 27: Please add CHECK_FEATURE call in qgl.cpp.
It is confusing when game uses many extensions, but does not report them to console.

renderer/Image_load.cpp
fholger: You're probably right. I took the formula from the glTexStorage manpage (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexStorage2D.xhtml at the very bottom in the errors section), but yeah, ILog2 is already integer, so floor does nothing.

renderer/Image_load.cpp
fholger: Old code. Some of our generated dynamic textures (most prominently currentRenderImage as the main FBO render texture) uses an outdated way for generation via R_RGBA8Image, which in turn uses image->GenerateImage to generate a small dummy texture that is then later overridden when the texture is resized to the actual FBO size. However, glTexStorage makes the texture storage immutable, so you can no longer resize the texture and thus the entire FBO setup is borked.

Yes, this code should be updated - render textures should not be following this code path. But I'm not sure if it's as easy as changing the generator from R_RGBA8Image to FB_RenderTexture, so I didn't want to do this here. Guess I should make a follow up ticket...

idlib/ParallelJobList.cpp
fholger: That's the question. If I don't silence it, we will get useless spam from the image load process. id probably added that log because it was a good indicator for how they were using the job system in D3 BFG. But currently we're only using it for two things - image loading during level load and adding model surfaces in parallel during frontend frame preparation. The warning is useless for the first use case, it might have merit for the second, I'm not sure. I guess we could either remove it entirely or somehow add an exception for the image load.

renderer/Image_load.cpp
stgatilov: Ok, I see.

renderer/Image_load.cpp
stgatilov: Ok. I did not think someone changes image size after creation.

idlib/ParallelJobList.cpp
stgatilov: I think that reflects the generic usage of job system in games: processing tasks which happen between two frames. It is not clear how much of that would happen in TDM. We can leave it as is, but if more jobs appear, then add settings about that in job system interface (aka "realtime" or "offline").

fholger: I actually found a better way - digging through the code again, id already added an exception to this rule: they have a third type of job queue for "utility" jobs that we haven't been using so far. That one is exempt from the warnings. I modified the image loading code to create its own utility job queue instead of recycling the frontend job list (which wasn't technically correct, anyway), so now the warnings are gone without having to increase the timeout value :)

renderer/qgl.h
fholger: Done.

renderer/Image_load.cpp
fholger: Yeah, I was surprised, too. Took me a moment of debugging time to understand why using glTexStorage broke rendering :D

renderer/Image_load.cpp
fholger: I removed Floor from the equation. As expected, does not appear to make any difference :)

renderer/Image_load.cpp
fholger: I also added the cvar.

stgatilov (approved):
I don't have any more comments.
I suggest committing it to SVN, and probably updating exe to get feedback from other developers =)

fholger:
Committed to svn.

PR #27, opened 2021-05-21
Replace ingame profiling with Tracy profiler

fholger:


CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ cmake_dependent_option(WITH_TOOLS "Compile with Tools" ON "MSVC" OFF)
 set(GAME_DIR "${PROJECT_SOURCE_DIR}/../darkmod" CACHE STRING "The directory where the game assets are stored")
 option(COPY_EXE "Copy executable to GAME_DIR after build" ON)
 set(THIRDPARTY_PLATFORM_OVERRIDE "" CACHE STRING "If you build conan dependencies yourself with custom platform_name, pass platform_name string here")
+option(ENABLE_TRACY "Enable tracy profiling support. Needs to be activated ingame by enabling the cvar tdm_enable_tracing." ON)
```
stgatilov: I think cvar tdm_enableTracing or tdm_enable_tracing should be renamed to com_enableTracing.

tdm_XXX cvar name was the convention back when the engine was closed source. As the result, it is almost exclusively about TDM game code. Today we try to add cvar like ID did, and tracing thing is about engine for sure.

engine.vcxproj
```diff
@@ -759,6 +759,7 @@
     </ClCompile>
     <ClCompile Include="framework\Session.cpp" />
     <ClCompile Include="framework\Session_menu.cpp" />
+    <ClCompile Include="framework\Tracing.cpp" />
```
stgatilov, line 762: It is quite interesting whether it should be in idlib or the main project.

Putting it into the main project means we cannot have traces in idlib code. Might become a problem in theory. On the other hand, we can easily change it later.

sys/msvc/properties/_Common.props
```diff
@@ -11,10 +11,10 @@
   </PropertyGroup>
   <ItemDefinitionGroup>
     <ClCompile>
-      <PreprocessorDefinitions>WIN32;_WINDOWS;_CRT_SECURE_NO_DEPRECATE;_CRT_NONSTDC_NO_DEPRECATE;_ALLOW_KEYWORD_MACROS;NO_WARN_MBCS_MFC_DEPRECATION;ExtLibs=;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>TRACY_DELAYED_INIT;TRACY_MANUAL_LIFETIME;TRACY_ENABLE;TRACY_ON_DEMAND;TRACY_NO_BROADCAST;TRACY_ONLY_LOCALHOST;WIN32;_WINDOWS;_CRT_SECURE_NO_DEPRECATE;_CRT_NONSTDC_NO_DEPRECATE;_ALLOW_KEYWORD_MACROS;NO_WARN_MBCS_MFC_DEPRECATION;ExtLibs=;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
stgatilov: `TRACY_ENABLE` can only be in Release configuration.

As we know, Debug with Inlines won't build. And I guess we can omit it in Debug it that's easier.

fholger: Thanks for the reminder. I knew I forgot something :D

